PR #8, opened 2025-07-26
Add constraints from weak deps to the registry tests

### LINKED ISSUE #7: Support for weak dependencies in the test code
From casual browsing of e.g. the test code I don't see that weak dependencies and constraints are being added. For example in

https://github.com/StefanKarpinski/Resolver.jl/blob/4f371082e965372fd679fa9a2562f8a7ae34081d/test/registry.jl#L31-L36

it looks like the fields in https://github.com/JuliaLang/Pkg.jl/blob/a8e394b6155a978ad38b65ec80058f16ee8dfa9b/src/Registry/registry_instance.jl#L60-L64 are not handled.

A weak dependency + weak compat should add a constraint but not add a requirement that the package is installed.

This might just require a small modification to the ingestion of the data and the resolver is probably flexible enough to handle it?

KristofferC:
This ensures that weak compatibility also adds constraints so that we do not resolve package versions that are incompatible with our extensions

Not sure what the best way to add a test for this is.

Fixes https://github.com/StefanKarpinski/Resolver.jl/issues/7.

### DIFF
--- a/test/registry.jl
+++ b/test/registry.jl
@@ -34,6 +34,7 @@ function provider(
             for (r, c) in info.compat
                 v in r && mergewith!(∩, comp[v], c)
             end
+            hasfield(typeof(info), :weak_compat) &&
             for (r, c) in info.weak_compat
                 v in r && mergewith!(∩, comp[v], c)
             end
--- a/test/registry.jl
+++ b/test/registry.jl
@@ -34,6 +34,9 @@ function provider(
             for (r, c) in info.compat
                 v in r && mergewith!(∩, comp[v], c)
             end
+            for (r, c) in info.weak_compat
+                v in r && mergewith!(∩, comp[v], c)
+            end
         end
         foreach(sort!, values(deps))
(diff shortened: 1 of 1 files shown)


StefanKarpinski (approved):
Yes, this seems good. Thought I had done this already, but I guess not.

StefanKarpinski:
Passes locally, finally managed to kick off CI...

test/registry.jl
```diff
@@ -34,6 +34,10 @@ function provider(
             for (r, c) in info.compat
                 v in r && mergewith!(∩, comp[v], c)
             end
+            hasfield(typeof(info), :weak_compat) &&
```
KristofferC, line 37: Sure, but is there a strong reason to support pre LTS? I suggest just bumping the CI matrix.

StefanKarpinski: 🤷🏻‍♂️ if it's easy to support older versions, why not?

PR #9, opened 2025-07-26
Allow setting julia version(s) in the resolver tests

KristofferC:
Right now, the julia version is completely filtered out which makes the resolver test less representative of how it is used in Pkg. This PR allows one to pass the set of Julia versions one want to resolve for and add some tests.

### DIFF
--- a/test/registry.jl
+++ b/test/registry.jl
@@ -7,7 +7,7 @@ import Pkg.Registry: RegistryInstance, init_package_info!
 import Pkg.Types: stdlibs
 import Pkg.Versions: VersionSpec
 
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
So this code is pretty old. The new, good version of this is [bin/Registries.jl](https://github.com/StefanKarpinski/Resolver.jl/blob/main/bin/Registries.jl), which we should copy or share here.

KristofferC:
Ah, yes, that looks better

StefanKarpinski:
Yeah, I'm fairly certain that one's correct since I can generate identical manifests to Pkg.jl.

PR #19, opened 2025-12-02
filter yanked packages in `bin/resolve.jl`

### LINKED ISSUE #15: Yanked versions are not recognized
I would love to use Resolver.jl through the [julia-downgrade-compat](https://github.com/julia-actions/julia-downgrade-compat) action, but I noticed a problem of Resolver.jl, where it chooses a version, which is yanked leading to an `Unsatisfiable requirements` error. More details are in https://github.com/julia-actions/julia-downgrade-compat/issues/19.

ranocha:
Fixes #15

I added a test that fails with the current version and passes locally with the version from this PR. 

I also updated the version in `Project.toml` to make it easier to release a new version of this package.

@StefanKarpinski I would be happy if you could review this PR (and the other ones I opened).


JoshuaLampert (approved):
Could this please be merged, @StefanKarpinski? This is holding several people up from using Resolver.jl. I think this PR should be relatively easy to review.

PR #23, opened 2026-04-03
Add support for workspaces

maleadt:
I'm entirely unfamiliar with Pkg internals, but at least I verified this works. I'll leave it up to @KristofferC to point out any obvious issues.

x-ref https://github.com/JuliaLang/Pkg.jl/pull/3841

### DIFF
--- a/bin/resolve.jl
+++ b/bin/resolve.jl
@@ -500,6 +500,27 @@ else # generate a manifest
         )
         for (uuid, info) in info_map
     )
+    # add root project as fixed entry for workspace manifests
+    let root_uuid = env.project.uuid
+        if root_uuid !== nothing && hasfield(typeof(env), :workspace) && !isempty(env.workspace)
+            root_deps = Dict{String,UUID}()
+            for (name, uuid) in env.project.deps
+                uuid in keys(deps) && (root_deps[name] = uuid)
+            end
+            root_weakdeps = Dict{String,UUID}()
+            for (name, uuid) in env.project.weakdeps
+                uuid in keys(deps) && (root_weakdeps[name] = uuid)
+            end
+            deps[root_uuid] = PackageEntry(;
+                uuid = root_uuid,
+                name = env.project.name,
+                version = env.project.version,
+                path = ".",
+                deps = root_deps,
+                weakdeps = root_weakdeps,
+            )
+        end
+    end
     # create manifest and record project hash
     manifest = Manifest(; julia_version, deps)
     env.manifest = manifest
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
Took this for a spin against current `main` and the [Pkg workspace docs](https://pkgdocs.julialang.org/dev/toml-files/). Overall it works and is correct for the common case; there's one gap for monorepo-style workspaces.

**Rebase + verification**

- Rebases cleanly onto current `main` (which has since touched the same manifest-write region).
- Full test suite passes on both Julia 1.12 and nightly. The workspace code is guarded by `hasfield(typeof(env), :workspace) && !isempty(env.workspace)`, so it's inert for non-workspace projects and pre-1.12 Julia — good.
- The field accesses line up with Pkg 1.12 (`env.workspace :: Dict{String, Pkg.Types.Project}`, keyed by member Project.toml path).

**Correctness vs. Pkg itself**

For the common shape — a root package plus a bare sub-environment (test/docs/benchmarks style) — the manifest this produces is **byte-identical to what `Pkg.resolve()` writes**, down to the same `project_hash`. That's the primary "manage test/docs dependencies in one manifest" use case, and it's spot-on.

**The gap: member projects that are themselves packages**

For the monorepo case the docs call out (`projects = [..., "PrivatePackage"]`), a workspace member with its own `uuid` is not emitted as a manifest entry. Concretely, with a member package `MemberPkg` at `mem/`:

- `Pkg.resolve()` → manifest contains `[[deps.MemberPkg]]` with `path = "mem"`.
- this PR → identical manifest **except `MemberPkg` is missing** (its registry deps are still folded in and resolved, but the package itself isn't dev-linked).

Root cause: part 1 correctly filters member deps to `known` (registry + stdlibs) so it doesn't try to *resolve* a local member from the registry — but part 2 only ever writes the **root** as a `path = "."` entry, never the members. The member directory is available (it's the `env.workspace` key, discarded by `for (_, ws_proj)`), and any member with a non-`nothing` uuid should get a `path = <dir>` entry with inter-member deps wired up.

**Smaller notes**

- `[sources]` path entries aren't handled (probably fine for registry resolution).
- A member's `[compat] julia` is folded into the julia compat, but `resolve.jl` ignores project-level julia compat anyway (only `--julia` constrains it), so it's a no-op — consistent with root behavior, not a regression.
- No tests; now that `main` has the `bin/test/` harness, a workspace regression test there would be easy to add.

I'm putting up a follow-up PR that rebases this and extends part 2 to emit every uuid-bearing workspace member as its own `path` entry (so the monorepo case matches Pkg too), with a regression test. Will link it here.

StefanKarpinski:
Follow-up with the rebase + member-package extension + regression test is up as #28.

StefanKarpinski:
Superseded.

PR #26, opened 2026-07-08
ci: fix broken CI (macOS matrix version + yanked-packages test)

StefanKarpinski:
Two independent, pre-existing failures were turning all of CI red:

- The macOS matrix `include` entries set `os`/`arch` but not `version`. Matrix `include` entries don't inherit the base matrix keys, so `version` was null and julia-actions/setup-julia failed immediately ("Version input must not be null"). Give each macOS entry `version: '1'`.

- The "yanked packages" test wrapped `run(resolve.jl ...)` in `@test_nowarn`, but on a clean depot resolving legitimately prints Pkg's "Installed ..." progress to stderr, which `@test_nowarn` treats as a warning. Assert `success(...)` instead: we care that resolution succeeds and doesn't pick the yanked Compat v4.0.0, not that Pkg is silent.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,12 +24,17 @@ jobs:
         arch:
           - x64
           - x86
-        # macos-14 and newer (Apple M series) only support aarch64 architecture
+        # macos-14 and newer (Apple M series) only support aarch64 architecture.
+        # These entries introduce new os/arch combinations, so they must also
+        # specify `version`: matrix `include` entries do not inherit it, and a
+        # null version makes julia-actions/setup-julia fail immediately.
         include:
           - os: macos-latest
             arch: aarch64
+            version: '1'
           - os: macos-15-intel
             arch: x64
+            version: '1'
     steps:
       - uses: actions/checkout@v5
       - uses: julia-actions/setup-julia@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,6 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.8'
           - '1' # automatically expands to the latest stable 1.x release of Julia.
           - 'nightly'
         os:
--- a/bin/Registries.jl
+++ b/bin/Registries.jl
@@ -1,13 +1,46 @@
 module Registries
 
-export registry_provider
+export registry_provider, package_info
 
 import Base: UUID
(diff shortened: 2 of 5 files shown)


StefanKarpinski:
Pushed a follow-up commit (`108e5eb`) fixing the failures that were still red once the macOS/`@test_nowarn` issues were out of the way — the two ends of the version matrix, `1.8` and `nightly`.

## Julia 1.8 — dropped

The core tests passed on 1.8, but the `bin/` tooling requires Julia ≥ 1.9 (its `Project.toml` is `julia = "^1.9"`, manifests start at v1.9), so the yanked-packages test (which shells out to `bin/resolve.jl`) couldn't pass there. 1.8 is long EOL, so it's removed from the matrix.

## Julia 1.14 (nightly) — Pkg internals moved

Nightly hid a cascade of Pkg-registry-internals changes, each only visible after the previous was fixed. Metaprogrammed around them so both stable and nightly pass:

1. **`init_package_info!` signature.** The one-arg `init_package_info!(::PkgEntry)` was removed in favor of `init_package_info!(::RegistryInstance, ::PkgEntry)`. Added a `package_info` helper that maps a `PkgEntry`'s `registry_path` back to its `RegistryInstance`; `test/registry.jl` gets the same treatment.

2. **PkgInfo is now uuid-keyed.** The parsed registry data changed representation:
   - `deps` value: `Dict{String,UUID}` → `Set{UUID}`
   - `compat` value: `Dict{String,VersionSpec}` → `Dict{UUID,VersionSpec}`

   `bin/Registries.jl` now reads either form and builds uuid-based data directly; the manifest path in `resolve.jl` reconstructs names from a uuid→name table (registered packages + all stdlibs + julia); `test/registry.jl` recovers names via the registry.

3. **Yanked-packages test on a prerelease Julia.** It resolved for the running Julia's `major.minor`, but `1.14` has no registered release to resolve against. On a prerelease it now targets the LTS and reads the resolved version with `--print-versions` instead of generating a manifest — full manifest generation (`download_source`/`fixups_from_projectfile!`) requires the host and target Julia to match, which can't hold on nightly. On released Julia the test is unchanged (still exercises the full manifest-generation path).

(1) and (2) are genuine Julia-1.14 compatibility fixes for the tooling, worth having regardless of CI; (3) is a test adjustment.

Verified locally: full suite green on both nightly (1.14.0-DEV) and the current stable release.

PR #30, opened 2026-07-12
UNSAT diagnostics: explain unresolvable requirements and list verified fixes

StefanKarpinski:
## What

When a resolution problem is unsatisfiable, Resolver.jl previously reported a bare `Unsatisfiable`. This PR adds a `diagnose` entry point that explains *why* resolution failed and lists *verified fixes*, and wires it into `bin/resolve.jl` so the command line diagnoses automatically on failure:

```julia
diagnose(data, reqs; compat, holds, max_fixes = 8) :: Diagnosis
```

Sample output (`show(io, MIME("text/plain"), diagnosis)`) for a case where `A` needs `C` 1 and `B` needs `C` 2, but only one `C` can be chosen:

```
Conflict 1: A, B cannot be satisfied together.
  • you require A
  • you require B
  • A (all versions) requires C at 1
  • B (all versions) requires C at 2

Verified fixes:
  1. drop requirement B
     → allows: A 1, C 1
  2. drop requirement A
     → allows: B 1, C 2

Upstream fixes:
  • a release of A relaxing its compat on C
    → allows: A 1, B 1, C 2
  • a release of B relaxing its compat on C
    → allows: A 1, B 1, C 1
```

Each verified fix is one or more actions — **drop requirement X**, **relax your compat on X**, **unpin X** — plus **upstream** suggestions (a dependency release with looser compat). Independent conflicts are reported separately, and a single fix combines one action per conflict (joined with "and"). The `allows:` line shows only the versions relevant to the conflict, not the whole resolved manifest.

`Diagnosis` is a structured object (conflicts with fact chains, fixes with resulting versions, upstream suggestions), so Pkg can render its own presentation; the `show` method is the default renderer.

## How

Everything is built on a separate, diagnosis-only SAT instance in which every *blameable constraint group* is guarded by a selector variable, toggled via picosat assumptions:

- **R(p)** — "p is required" (one per direct requirement)
- **C(p)** — the user's compat/pin on p (received as data, not pre-filtered, so it can be blamed and relaxed)
- **H(p)** — a manifest hold / pin of p at a specific version
- **B(p, q, versions)** — a third-party compat bound, coalesced so one group ≈ one compat line in one Project.toml

On top of that, three analyses (all verified by construction, since every artifact comes off a SAT model or an unsatisfiability certificate):

1. **How many stories:** requirement-level MICE partition — each disjoint minimal unsatisfiable subset of requirements is one independent conflict, reported separately.
2. **Which story:** one group-MUS per conflict, shrunk with a biased deletion order (bounds first, user-owned constraints last) so that when a conflict *can* be explained through constraints the user owns, it is. Every fact in the chain is load-bearing: removing any one of them dissolves the conflict. Shrinking is a single linear pass — monotonicity of satisfiability under assumption removal makes restart scans provably redundant.
3. **Which fixes:** greedy MCS enumeration over user-owned selectors (keep-order: requirements, then pins, then compat), so "relax your compat" fixes surface before "drop your dependency" fixes. Each fix is a minimal correction set, globally verified (it repairs *all* conflicts at once), and reported with the concrete versions you'd get. Blocking clauses force each successive fix to be undominated.

Upstream suggestions probe each third-party bound in a conflict's MUS individually (per-cluster): if disabling just that bound makes the conflict satisfiable, a release relaxing it would resolve the conflict.

Reported solutions are pruned to the dependency-reachable closure of their satisfied requirements (SAT models may set unconstrained package variables true).

## Command line (`bin/resolve.jl`)

`bin/resolve.jl` resolves a real project against the registries and, when resolution fails, now prints the same diagnosis automatically before exiting nonzero — replacing the old bare `error("Unsatisfiable")`. UUIDs in the report are rendered as package names. Two new options make conflicts reproducible and expressible:

- `--registry=<path>` — resolve against a specific registry (e.g. a checkout of General at a pinned commit) instead of the installed ones.
- `--pin=<pkg@version>` — pin a package to an exact version (repeatable, comma-separated; leading `v` optional). Pins constrain resolution like a `Pkg.pin`, and are surfaced by the diagnosis as an `unpin` fix.

Real example, reproducible against General at [`6bb1cf95…`](https://github.com/JuliaRegistries/General/commit/6bb1cf9538b1db31c2f4e4517ede70bc4c44227e):

```console
$ julia --project=bin bin/resolve.jl myproject --julia=1.11 --registry=./General --pin=HTTP@0.6.10
Unresolvable. Diagnosis:

Conflict 1: AnthropicClient cannot be satisfied.
  • you require AnthropicClient
  • AnthropicClient (all versions) requires HTTP at 1.0.0–1.11.0
  • HTTP is pinned at 0.6.10

Verified fixes:
  1. unpin HTTP
     → allows: AnthropicClient 0.1.0, HTTP 1.0.0
  2. drop requirement AnthropicClient
     → allows: nothing
```

## Docs

`docs/unsat-diagnostics.md` documents the feature: how it works (MUS/MCS, the fix kinds), synthetic examples with canned data covering every fix kind and combined multi-conflict fixes, and reproducible real-registry examples invoked through `bin/resolve.jl` against a pinned General commit.

## Notes

- The `resolve` fast path is untouched; `diagnose` builds its own instance from unfiltered `PkgData` plus explicit user constraints.
- The `diagnose` test suite is 208 tests (plus a CLI smoke test), including a property test that sweeps all tiny UNSAT cases and independently re-verifies every fix, MUS minimality, and solution closure-tightness.

## Co-author

Implemented with [Claude Code](https://claude.com/claude-code): design, implementation, CLI integration, docs, and review by Claude, directed and reviewed by @StefanKarpinski.

https://claude.ai/code/session_01BC5icYSdvJZoU7FYsjZm5p

### DIFF
--- a/bin/resolve.jl
+++ b/bin/resolve.jl
@@ -17,9 +17,14 @@ usage: $PROGRAM_FILE [options] [<project path>]
   --julia=<versions>      Julia versions to resolve for (default: 1+)
                           use registry compat syntax, not semver
 
+  --registry=<path>       resolve against the registry at <path>
+                          (default: the installed registries)
+
   --allow-pre[=<pkgs>]    allow prerelease versions
   --extra-deps=<pkgs>     extra packages to require
   --prioritize=<pkgs>     package names/uuids to prioritize
+  --pin=<pkg@version>     pin a package to an exact version (repeatable,
+                          comma-separated; leading 'v' optional)
 
   --fix[=<pkgs>]          prefer current full version number
   --fix-minor[=<pkgs>]    prefer current major.minor version
@@ -48,7 +53,7 @@ Wherever <pkgs> appears you can specify a comma-separated list of:
 
 parse_opts!(ARGS, split("""
     print-manifest print-versions
-    julia allow-pre extra-deps prioritize
+    julia registry allow-pre extra-deps prioritize pin
     fix fix-minor fix-major unfix
     max max-minor max-major
     min min-minor min-major
@@ -90,7 +95,7 @@ end
 import Pkg.Registry: JULIA_UUID, PkgEntry, RegistryInstance,    init_package_info!, reachable_registries
 import Pkg.Types: Context, EnvCache, Manifest, PackageEntry, get_last_stdlibs, write_manifest
 import Pkg.Versions: VersionSpec, semver_spec
-import Resolver: Resolver, DepsProvider, PkgData, resolve
+import Resolver: Resolver, DepsProvider, PkgData, resolve, diagnose
 import HistoricalStdlibVersions: STDLIBS_BY_VERSION, UNREGISTERED_STDLIBS
 import TOML
 
@@ -159,11 +164,18 @@ let proj = env.project
     end
 end
 
-## load packages from installed registries
+## load packages from the registries
+
+# --registry=<path> resolves against a specific registry (e.g. a checkout of
+# General at a particular commit, for reproducibility); default is whatever
+# registries are installed in the depot
+const registries = handle_opts(:registry, nothing) do path::String
+    RegistryInstance(path)
+end |> r -> isnothing(r) ? reachable_registries() : [r]
 
 const packages = Dict{UUID,Vector{PkgEntry}}()
 
-for reg in reachable_registries()
+for reg in registries
     for (uuid, entry) in reg.pkgs
         push!(get!(()->PkgEntry[], packages, uuid), entry)
     end
@@ -291,6 +303,24 @@ global const reqs = copy(project_deps)
 
 handle_opts(:extra_deps) do val::String
     union!(reqs, parse_packages(val))
+end
+
+## options: pinned versions
+
+# --pin=<pkg@version>[,...] pins packages to exact versions (repeatable). Pins
+# constrain resolution (like a `Pkg.pin`) and, when a resolve is unsatisfiable,
+# are surfaced by the diagnosis as pins that could be lifted.
+const pins = Dict{UUID,VersionNumber}()
+
+handle_opts(:pin) do val::String
+    for item in split(val, ',')
+        m = match(r"^\s*(.+?)@v?(\d[^\s]*)\s*$", item)
+        isnothing(m) && usage("Invalid --pin (expected pkg@version): $item")
+        uuid = only(parse_packages(m[1]))
+        ver = tryparse(VersionNumber, m[2])
+        isnothing(ver) && usage("Invalid version in --pin: $item")
+        pins[uuid] = ver
+    end
 end
 
 ## options: sorting versions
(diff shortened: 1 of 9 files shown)


StefanKarpinski:
@KristofferC do you think this would be good enough to replace the old resolver?

KristofferC:
I did a bit of investigation, and w.r.t performance, it is typically ok, there are some cases where it is a lot slower, but fortunately that doesnt seem to originate from PicoSAT itself and they seem manageable by Claude (will open some PRs). Wr.t the unsat logs it is a bit hard to say because it depends a lot from person to person what they like, but we at least we now have a proper shot at integrating it. I would probably not delete the old resolver straight away but use some feature flag thing (probably with the new one as the default so it gets properly tested).

KristofferC:
I tried to stress test this PR a bit and hit some bad scaling where the diagnosis never finished. I asked Claude to write up what it found:

----------

:robot: 

Some field data from running `diagnose` against real General-registry problems — inputs captured from Pkg's own resolver pipeline (`deps_graph` output converted to `PkgData`, julia modeled as a pinned single-version package).

**output quality is excellent**. On a Makie+Plots+Images project with `ColorTypes = "0.9"` compat, the diagnosis names the correct fix first ("relax your compat on ColorTypes") in 63 lines, where Pkg's current resolver produces a 161-line restriction tree with the cause buried. Conflict clustering correctly separated three independent conflicts in one project into three stories.

**Scaling is the main issue.** `diagnose()` builds `DiagSAT` over whatever `data` you hand it, with no pruning — unlike the resolve path, which runs `filter_pkg_info!` (reachability + redundancy) before building its SAT. Three costs stack up: the selector-guarded bound groups generate `|p_versions| × |excluded_versions|` ternary clauses per compat declaration, which explodes on registry-scale data; `group_mus` is a linear deletion pass — one incremental SAT solve per selector group, so thousands of solves on a huge instance; and fix enumeration adds more solves on top. Measured wall times (diagnosis only, warm):

| data size (pkgs / versions) | time |
|---|---|
| 104 / 1.4k | 0.4 s |
| 203 / 5.2k | 10 s |
| 218 / 5.6k (3 clusters) | 26 s |
| 660 / 15k | 54 s |
| ~1474 (Pkg's weakdep-inflated closure) | DNF, killed at ~15 min |

The last row is partly on my harness — Pkg's `deps_graph` discovers through weakdeps, so its closure is much bigger than what `pkg_data`'s strong-dep walk produces — but the 660-pkg row is a perfectly ordinary Makie-sized project via the native closure. Concrete remedies: collapse versions with identical dep/compat footprints into equivalence classes before building the SAT (registry data is highly redundant; the provider already dedups the raw dicts), and/or QuickXplain-style dichotomic MUS extraction (O(k·log n) solves for a size-k MUS instead of O(n)).

**Smaller output nits, all presentational:**

- *Fix witnesses*: `enumerate_fixes` verifies each fix by solving and reports whatever model picosat returns — with no optimization that's often ancient versions ("relax your compat on ColorTypes → allows: Makie **0.11.1**, Plots **1.1.0**"), which reads as a recommendation to install 2020 software. Running the optimizing resolve on the fixed problem would produce a truthful witness (or label it "e.g.").
- *Upstream fixes*: one suggestion per bound in the MUS whose removal helps — the tangle above produced 10, including "a release of BinaryProvider relaxing its compat on julia". Needs ranking/capping.
- *Story tangents*: the MUS is minimal in the formal sense but can include facts that exist only to rule out ancient versions (the chain above detoured through StatsMakie/FillArrays); the rendering presents them at the same level as the two lines that matter.
- One API thought: a non-negotiable-requirement class would help integrators. I modeled julia as a required package and got "drop requirement julia" offered as a fix; `bin/resolve.jl` avoids this by modeling julia as a dependency edge instead, but both modelings should be able to say "this requirement can appear in explanations but never in fixes".

StefanKarpinski:
To clarify, was the slowness in diagnostics or in solving satisfiable problems or both? I had tested solving satisfiable problems that were quite large, but there may be cases that cause some pathological amount of backtracking or something. The main solver already does aggressive deduplication of redundant versions, the diagnostic solver doesn't. I suspect the best approach to speeding up the diagnostic solver is using a hybrid approach where only causes that may need to be in fixes are treated that way and then we handle everything less the way the main solver does. It's also possible that we can do the diagnostics with the reduced problems; I'm not actually convinced that we can't.

KristofferC:
Both, the diagnostics was the bad scaling, the performance was mostly fixed by the two PRs I opened. The Pkg resolver has a fast path that is very fast (and quite often "hits") and in those cases it can be quite a bit faster. It is possible to add such a path to this package as well but the solve time is starting to be low enough in absolute terms that it perhaps doesn't matter.

KristofferC:
https://github.com/JuliaLang/Pkg.jl/pull/4738, it uses this PR + #34. Here is an example that takes a long time for the diagnostics:

```
julia> @time try Pkg.add([(;name="DataFrames", version="1.7"), (;name="PrettyTables", version="1")]) catch end
   Resolving package versions...
 10.375237 seconds (5.20 M allocations: 869.659 MiB, 4.72% gc time)
```

the message is

```
  • you require DataFrames
  • your compat restricts DataFrames to 1.7.0–1.7.1
  • DataFrames 1.4.0–1.7.1 requires PrettyTables at 2.1.0–2.4.0
  • your compat restricts PrettyTables to 1.0.0–1.3.1

Verified fixes:
  1. relax your compat on DataFrames
     → allows: PrettyTables 1.0.0, julia 1.14.0-DEV.2787, Plots 1.10.3, DataFrames 1.0.0
  2. relax your compat on PrettyTables
     → allows: PrettyTables 2.1.0, julia 1.14.0-DEV.2787, Plots 1.10.3, DataFrames 1.7.1
     → allows: PrettyTables 1.0.0, julia 1.14.0-DEV.2787, Plots 1.10.3

Upstream fixes:
  • a release of DataFrames relaxing its compat on PrettyTables
    → allows: PrettyTables 1.0.0, DataFrames 1.7.1
```

Some questions: why does `Plots` end up in the trace? The `julia` part seems like noise? The "compat" term doesn't seem completely accurate, here I used specific restricted versions with the `@` syntax, it wasn't compat from my project file. 

The maxsum log is:

```
julia> Pkg.add([(;name="DataFrames", version="1.7"), (;name="PrettyTables", version="1")])
   Resolving package versions...
ERROR: Unsatisfiable requirements detected for package DataFrames [a93c6f00]:
 DataFrames [a93c6f00] log:
 ├─possible versions are: 0.11.7 - 1.8.2 or uninstalled
 ├─restricted to versions 1.7 by an explicit requirement, leaving only versions: 1.7.0 - 1.7.1
 └─restricted by compatibility requirements with PrettyTables [08abe8d2] to versions: [0.11.7 - 0.21.8, 1.0.0 - 1.3.6] or uninstalled — no versions left
   └─PrettyTables [08abe8d2] log:
     ├─possible versions are: 0.1.0 - 3.4.2 or uninstalled
     └─restricted to versions 1 by an explicit requirement, leaving only versions: 1.0.0 - 1.3.1
```

KristofferC:
Tell me when I should do another Pkg test drive. Maybe not worth doing until the diagonstic perf gets a bit better.

StefanKarpinski:
Yeah, I'll let you know — it's almost there. Had to do a bunch of fairly deep refactoring to make it so the regular filtered and efficient SAT instance can also be used for diagnostics.

StefanKarpinski:
Superseded by #57, which rebuilds this feature from scratch on the architecture that landed in the meantime (#49 Problem/constraints-as-data, #50 universal T1, #51 stdlib semantics) — diagnosis now runs on the same filtered SAT instance resolve uses, which eliminates the scaling problem reported here at the root (the 10.4s/54s/DNF cases measure 1.25s/0.72s/1.47s there). Every review point from this thread is addressed point-by-point in #57's description. @KristofferC — this would be the build for the next Pkg test drive.

PR #32, opened 2026-07-23
perf: specialized single-solution descent for max == 1

KristofferC:
The enumeration machinery needs nested temp scopes and a re-solve per
dependency layer so it can rewind and search for further solutions;
when only the first solution is wanted neither is necessary. Add
find_first_solution: an iterative descent sharing the per-layer
optimization body with the enumerating search (optimize_opts!,
next_opts, save_solution!).

Keeping a picosat context open for the whole search measurably
degrades solving, so with the new `restore=false` keyword the descent
runs outside any temp scope and leaves the instance constrained by the
requirements and the solution's pins. The default `restore=true`
preserves the reusable-instance contract that callers like
slices/Slices.jl rely on (locked in by a new test); the info-level
resolve methods own their single-use instance and opt in.

2.5-6.5x on real General-registry resolves on top of the previous
commit (mega 6-package env 4.2 s -> 0.65 s total since baseline).
Identical solutions on all five registry captures and on 500 randomized
tiny instances (both max=1 and full enumeration); test suite passes.

### DIFF
--- a/src/Resolve.jl
+++ b/src/Resolve.jl
@@ -6,10 +6,55 @@ function resolve_core(
     reqs :: SetOrVec{P} = keys(sat.info);
     max  :: Integer = DEFAULT_MAX_SOLUTIONS,
     by   :: Function = identity, # ordering
+    restore :: Bool = true, # restore the SAT instance's state before returning
 ) where {P}
     # solution search data structures
     sol = Dict{P,Int}() # current solution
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
I actually think we should change the API to only give a single solution. The idea behind multiple solutions was that it would help with diagnostics in unsat cases, but the diagnosis system didn't end up using that at all.

PR #36, opened 2026-07-28
Front-necessity resolver (parked reference implementation)

StefanKarpinski:
**Status: parked reference implementation — not intended to merge.** This is the working implementation of the *front-necessity* semantics analyzed in the theory docs (#35): resolve by pinning, in priority order, the packages contained in every Pareto-optimal tight solution consistent with the pins, at the best version among those solutions.

It exists as a PR so the diff is reviewable and so the theory documentation's reference to the `front-necessity` branch resolves. The full analysis — the semantics, Theorems A–D, the counterexamples, and the reasons this direction was set aside — is in `docs/design/front-necessity.md` on this branch and, in polished form, in the Theory section added by #35.

## What's here

- `resolve` implements the forced descent over the ⊵-front: candidates are constructed by the dependency-layered greedy under per-query assumptions (junk-free greedy answers are provably Pareto-maximal in their query space); all queries range over supported models; processed tight cores are excluded with their entire junk-variant families by one clause each; certified front members are cached and certify most pins with no solves; pins are carried as solver assumptions.
- The brute-force test reference implements the definition verbatim on raw data (enumerate tight solutions, form the front, run the descent); the implementation matches it exactly on ~19,500 instance/order pairs, and the full test suite passes.
- Data-level entry points build the **unreduced** problem: the reachability filter provably drops front members, and Theorem D shows no version-level reduction can be faithful to necessity.

## Why it's parked

Necessity and dominance are universal queries; they can't run on reduced problems, and the costs are intrinsic: small resolves are fast (JSON ~4 ms, DataFrames ~85 ms on unreduced instances), but `DifferentialEquations` (761 packages / 26k versions unreduced) is intractable. The layered semantics achieves the same decoupling goal — a canonical filter-free definition with provably-safe reduction — at none of the cost (see the theory docs' reduction-invariance theorem). The `DifferentialEquations`-scale registry test cases are commented out on this branch pending a front-preserving reduction, which is the open problem recorded in the docs.

## Co-author

Written with [Claude Code](https://claude.com/claude-code); design and proofs developed interactively with @StefanKarpinski.

### DIFF
--- a/docs/design/front-necessity.md
+++ b/docs/design/front-necessity.md
@@ -1,0 +1,296 @@
+# Necessity on the Pareto front: canonical single-solution semantics
+
+Status: design notes for the `front-necessity` branch. Builds on the
+single-solution API (`resolve` returns a package ⇒ version `Dict`, or
+`nothing` when unsatisfiable). Complements issue #33.
+
+## 1. Motivation
+
+The single-solution resolver pins packages one at a time, best version
+first, in a caller-supplied priority order. Two questions determine its
+semantics: *which package gets pinned next*, and *relative to what set of
+solutions "best" and "next" are judged*.
+
+The dependency-layered descent answers the first question syntactically
+(requirements first, then dependencies of chosen versions, layer by
+layer), which subordinates the priority order to dependency depth and ties
+the resolver to the dependency structure. The natural alternative —
+repeatedly pin the highest-priority package that is *necessary*, i.e.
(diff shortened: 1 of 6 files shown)


StefanKarpinski:
Opened by misunderstanding — the docs PR is #35. Keeping the branch pushed since the theory docs reference it as the parked reference implementation.

PR #38, opened 2026-07-28
Filter machinery optimizations: 15× faster filtering, plus pkg info file fix

StefanKarpinski:
A sequence of single-fix performance commits to the filtering machinery (each one passes the full test suite on its own), plus one bug fix found along the way. End result on the DifferentialEquations registry closure (764 packages / 26,361 raw versions), best of 5:

| stage | before | after | speedup |
|---|---|---|---|
| reach | 557ms | 16ms | 35× |
| redundancy | 175ms | 23ms | 7.7× |
| drop | 14ms | 9ms | 1.6× |
| **filter total** | **747ms** | **48ms** | **15.6×** |
| unfiltered info build | 182ms | 105ms | 1.7× |
| resolve end-to-end | 1188ms | 411ms | 2.9× |

The requirement-independent dominance-first split lands at 34ms for the cacheable pass + 12.5ms per resolve. Small instances (JSON, DataFrames) stay at or under a millisecond.

### How

- **Word-aligned column layout**: every conflicts BitMatrix pads its row dimension to whole 64-bit words, so each column is an aligned chunk span. Since conflicts are stored symmetrically, every pairwise scan has an orientation in which it's a contiguous column read — reach scans the partner-side column with one masked word scan, and redundancy elimination becomes word-parallel column intersection instead of per-bit row subset tests. The pkg info file format inherits the padded geometry (magic bumped to v2).
- **Package→integer interning** in both fixpoints: no name hashing in inner loops; reach's queue becomes a stack + pending vectors, redundancy's work set becomes flag bits, and the `Iterators.cycle` rescan is replaced by a FIFO queue (O(1) wake-ups).
- **Batched dominance**: per-version candidate masks updated in one sweep over active columns, with versions dropping out as their candidates empty; identical adjacent columns (ubiquitous, since compat bounds are ranges) are skipped outright.
- **drop_unmarked!** computes kept-column masks directly (partner activity word-blitted via a new `copy_col_bits!` kernel) instead of round-tripping through strided flag-row writes, and rebuilds matrices by word copy (prefix fast path) or chunk gather.
- **Build**: conflict masks computed once per distinct compat entry (providers deduplicate compat dicts across versions) and reused by both build passes, with distinct entries found by identity scan; partner-side conflict columns blitted word-wise; shared dependency vectors deduplicated by identity before unioning; remaining strided single-bit writes lowered to direct chunk stores.
- **Fix**: `save_pkg_info_file`/`load_pkg_info_file` were broken for non-String package types (stringified index lookups, no Symbol reconstruction); fixed with roundtrip tests for String- and Symbol-keyed data.

### Verification

Every optimization is behavior-preserving and verified as such: reach maps, filtered versions, and answers are **byte-identical** to the pre-optimization implementation across a 200k-instance differential dump (fixed and version-varying dependency generators) and registry checksums. Where processing order changes (FIFO queue, batched dominance), the commit messages include the confluence argument for why the fixpoint is order-independent. Full test suite passes at every commit.

### DIFF
--- a/src/BitKernels.jl
+++ b/src/BitKernels.jl
@@ -24,18 +24,6 @@ function col_copy!(buf::Vector{UInt64}, X::BitMatrix, j::Int)
     return buf
 end
 
-# buf .&= column j of X; returns true if buf has any set bit left
-function col_and!(buf::Vector{UInt64}, X::BitMatrix, j::Int)
-    W = col_words(X)
-    base = (j - 1) * W
-    chunks = X.chunks
-    live = UInt64(0)
-    @inbounds for w = 1:W
-        live |= (buf[w] &= chunks[base + w])
-    end
-    return !iszero(live)
-end
-
 # does column j of X intersect the row set in buf?
 function col_intersects(X::BitMatrix, j::Int, buf::Vector{UInt64})
     W = col_words(X)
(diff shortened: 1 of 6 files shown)


StefanKarpinski:
(This landed after the merge — the commit described here is now PR #39 instead.)

PR #39, opened 2026-07-28
Run redundancy elimination first and iterate the filter to a fixpoint

StefanKarpinski:
Follow-on to #38 (item 5 of the pre-SAT pipeline improvements): `filter_pkg_info!` now runs redundancy elimination *before* reachability — it needs no reachability marks, and on registry-scale problems it does most of the shrinking (the DifferentialEquations closure drops 26,361 → 5,857 versions before reach even runs) — and then alternates the two passes until neither deletes anything. Requirement packages always survive filtering, and each effective round strictly shrinks the total version count, so the loop terminates. Every round preserves the layered answer (Theorem 4); the theory docs' "Iterating the filter" section is updated to describe the new schedule.

This is the first output-changing commit of the series — kept sets can only shrink:

- Across the 200k-instance differential dump: answers and raw reach maps **byte-identical** to the previous implementation; kept sets strictly smaller in 1,763 cases, larger in none (subset-checked).
- DiffEq closure: keeps drop from 5,408 versions / 732 packages to **5,219 / 702**.
- The filter itself costs ~10ms more (the dominance pass sees the raw problem), but the smaller SAT encoding more than pays for it: resolve end-to-end 491ms vs 502ms (best of 10, measured back-to-back).

This ordering is also the split that future registry caching wants: the opening redundancy pass is requirement-independent and cacheable per registry state, leaving only the ~12ms iterate-to-fixpoint tail per resolve.

Full test suite passes; docs build clean.

### DIFF
--- a/docs/src/theory/layered.md
+++ b/docs/src/theory/layered.md
@@ -235,10 +235,10 @@ layered solution.**
 
 ## Iterating the filter
 
-The filter runs reachability once and redundancy elimination once, in
-that order. It is natural to ask whether iterating the two passes would
-prune more — and the answer illuminates what each pass can and cannot
-see.
+It is natural to ask whether a single reachability pass followed by a
+single redundancy pass leaves anything on the table — whether iterating
+the two would prune more. The answer illuminates what each pass can and
+cannot see.
 
 Reachability alone is a closure operation:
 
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
Review question follow-up: measured whether the loop's repeated `drop_unmarked!` calls (which rebuild the package bitmaps) are worth avoiding. Three schedules, all producing byte-identical output (verified on the 200k-instance dump and the registry closure):

1. **Intermediate drops** (as committed): drop after the initial redundancy pass and after every round.
2. **Single drop at the end**: all passes iterate on the activity flags only. Requires reach to skip deactivated versions, conflict scans masked by partner activity, and clear-only `mark_reachable!` (so reachability can't resurrect versions redundancy killed).
3. **Hybrid**: drop once after the initial redundancy pass (the 26,361 → 5,857 shrink), then flag-only iteration and one final drop.

Interleaved A/B on the DifferentialEquations closure (filter time, best of 5, three runs each):

| schedule | filter | resolve end-to-end |
|---|---|---|
| intermediate drops | 57.4 / 57.9 / 59.0 ms | 481–507 ms |
| single drop at end | 64.2 / 64.6 / 65.9 ms | 480–497 ms |
| hybrid | 57.7 / 57.9 / 58.1 ms | 488–497 ms |

Why the rebuilds don't cost what one would expect: `drop_unmarked!` has an everything-active fast path per package, so the fixpoint-confirming round's drop rebuilds nothing, and effective rounds rebuild only packages that actually shrank, at cost proportional to the shrunken size. Meanwhile physically dropping is what makes later rounds cheap — the pure single-drop variant runs every round on the full 26k-version-wide matrices and loses ~13% on the filter; the hybrid breaks even but needs extra machinery (masked scan kernel, clear-only reach semantics, skip-inactive checks) purely to tie.

Conclusion: keeping the committed schedule — drops self-elide when there's nothing to drop and pay proportionally when there is, with the simplest pass invariant (every pass starts from freshly dropped or all-active matrices).

PR #41, opened 2026-07-28
Installability filtering: delete versions that appear in no model

StefanKarpinski:
Follow-on to #39 (extends its filtering loop) — the model-free version deletion pass, the one *semantic* pre-SAT improvement from the earlier analysis: a version one of whose dependencies has no surviving compatible version appears in **no model at all** — any model containing it would have to assign that dependency some compatible version. Deleting such versions changes nothing about the model set; it's the strongest license any filter rule can have (stronger than answer preservation). The pass (`mark_installable!`) is arc-consistency on the dependency constraint, iterated to a fixpoint since each deletion can strand more versions.

Implementation notes:

- Word-parallel: dependency `q` is live for `p@i` iff `q`'s active-version mask escapes the contiguous conflict column `Y[·, c+i]` (one word-scan per dep check), with a package-level reverse-dependency work queue driving the fixpoint.
- Runs both before the filtering loop (it's requirement-independent, so it belongs with the cacheable dominance pass) and inside each round (reach/redundancy deletions can strand more versions).
- New edge case handled: installability is the only pass that can delete a *required* package's every version. That means the problem is unsatisfiable — reach is now seeded from the surviving requirements, and `resolve` returns `nothing` instead of erroring on the missing package.
- Theory docs updated: the Filtering section describes the third pass and its trivial license (with the induction argument that no model's version is ever deleted); new unit testset covers the simple case, a cascade that empties the requirement (unsat), and a surviving alternative version.

Verification: across the 200k-instance differential dump, answers and raw reach maps are **identical** to #39's implementation while kept sets shrink in **55,247 of 200k cases** (always strict subsets — the random generators produce many broken dependency cones). On the DifferentialEquations closure the effect is small (79 raw versions deleted in 3.5ms; final keeps 5,219 → 5,188; resolve time unchanged within noise) — the General registry's reachable cones are mostly intact. The payoff is on problems with stale compat or yanked versions, where whole broken subtrees now vanish before the SAT encoding sees them, and in the future cached-registry design, where this pass joins dominance in the requirement-independent cacheable stage.

Full test suite passes; docs build clean.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -8,4 +8,5 @@ resolve
 
 ```@docs
 Resolver.find_reachable
+Resolver.mark_installable!
 ```
--- a/docs/src/theory/layered.md
+++ b/docs/src/theory/layered.md
@@ -54,7 +54,7 @@ itself, which pins the unique answer.
 
 ## Filtering
 
-Real problems are shrunk before solving by two passes (`pkg_info` with
+Real problems are shrunk before solving by three passes (`pkg_info` with
 `filter=true`, the default):
 
 **Reachability** (`find_reachable`) computes, per package, a version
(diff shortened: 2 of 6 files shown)


StefanKarpinski:
Closing after measurement — the pass doesn't carry its weight on real data.

The motivating hypothesis was that the registry is full of broken dependency cones (versions whose compat pins vanished/yanked versions of deps), so deleting provably model-free versions would meaningfully shrink the problem. Measured, that's false:

- DifferentialEquations closure, yanked versions **included** (test provider): installability deletes **79 of 26,362** raw versions (~0.3%), final keeps 5,219 → 5,188, resolve time unchanged within noise.
- Same closure with yanked versions **excluded** (as the bin tooling does — the strongest version of the hypothesis, since compat pinned to a yanked version becomes a broken cone): still only **79 of 26,083**.
- The 55k/200k random-instance shrinkage cited in the PR body is a generator artifact: the random tester frequently emits compat entries excluding *every* version of a dependency, which real registry data (AutoMerge compat checks, grow-only version lists) essentially never does.

Meanwhile the pass costs ~120 lines plus a new edge-case path that exists only because of the pass itself (a required package emptied by filtering — the SAT solver already handles broken cones correctly without any of this; the pass only prunes them earlier). Cheap provable hygiene, but not worth the complexity.

Keeping the branch as a reference in case a future data source actually exhibits broken cones at scale — the license argument (deletes only versions appearing in no model, by induction along the deletion order) and the word-parallel implementation both hold up; it's the impact that doesn't.

PR #43, opened 2026-07-28
Rectangle-encoded conflict clauses

StefanKarpinski:
Follow-on to #42 — implements the conflict-encoding compression measured there: conflicts come from compat ranges, so the pairwise encoding (535k binary clauses on the DifferentialEquations closure) is massively redundant.

**Encoding.** Within each interacting pair's conflict bitmap, p's versions are grouped by their conflict pattern against q — identical patterns are the norm since conflicts come from shared compat entries, and each pattern is read off a contiguous column of q's matrix. Each pattern splits into maximal runs of q's versions, and each (version group × run) rectangle becomes one clause `¬x ∨ ¬y` over trigger literals meaning "some version in the group/run is chosen": the version literal for singletons, the package variable when all versions are covered, otherwise a shared auxiliary variable defined by one implication per covered version (interval triggers shared globally by `(package, lo, hi)`, group triggers by content).

**Soundness.** Triggers occur only negatively, so one-directional definitions suffice: a genuine conflict forces both triggers true and violates the rectangle clause, while any conflict-free assignment extends by setting each trigger to its defining disjunction — models project bijectively onto the original variables, so satisfiability, the model set, and the layered answer are all unchanged. Verified: byte-identical 200k-instance differential dump, identical registry solutions, full test suite (including the 373k-case adversarial set and brute-force reference comparisons).

**Measured** (DifferentialEquations closure, interleaved best-of runs):

| | pairwise | rectangles |
|---|---|---|
| total clauses | 674k | 247k |
| conflict clauses | 535k | 11.2k rectangles + ~101k definitions |
| extra variables | — | 6.8k |
| SAT build | 60ms | 29ms |
| resolve end-to-end | ~355ms | ~279ms |

The end-to-end gain exceeds the build savings — the compact encoding propagates better, so the solves themselves speed up.

Remaining headroom noted for the record: the q-side interval definitions (~78k of the ~101k definition clauses) could shrink to ~3 clauses/version with shared prefix-ladder (order-encoding) variables, which would also replace the 87k pairwise mutex clauses with ~1 clause/version — but that requires biconditional ladders and touches the at-most-one structure, so it's left for a future round if the encoding is still hot.

### DIFF
--- a/src/SAT.jl
+++ b/src/SAT.jl
@@ -85,18 +85,108 @@ function SAT(
                 end
             end
 
-            # conflicts
-            #   ∀ q@j ∈ conflicts(p, i): !p@i OR !q@j
-            for i = 1:n_p
-                for (q, b) in info_p.interacts
-                    p < q || continue # conflicts are symmetrical
-                    n_q = length(info[q].versions)
-                    v_q = vars[q]
-                    for j = 1:n_q
-                        info_p.conflicts[i, b+j] || continue
-                        # conflicting versions are mutually exclusive
-                        PicoSAT.add(pico, -(v_p + i))
-                        PicoSAT.add(pico, -(v_q + j))
+            # conflicts are added below, rectangle-encoded
+        end
+
+        # conflicts, encoded as rectangles: within each interacting pair's
+        # conflict bitmap, group p's versions by their conflict pattern
+        # against q (identical patterns are the norm, since conflicts come
+        # from shared compat entries) and split each pattern into maximal
+        # runs of q's versions. each (version group) × (run) rectangle
+        # becomes one clause ¬x ∨ ¬y over trigger literals meaning "some
+        # version in the group/run is chosen": the version literal itself
+        # for singletons, the package variable when every version is
+        # covered, and otherwise a shared auxiliary variable defined by
+        # one implication per covered version. triggers occur only
+        # negatively, so the one-directional definitions suffice: a real
+        # conflict forces both triggers true and violates the rectangle
+        # clause, while every conflict-free assignment extends by setting
+        # each trigger to the value of its defining disjunction.
+        psets = Dict{Tuple{Int,Vector{Int}},Int}() # (v_p, versions) => var
+        qints = Dict{NTuple{3,Int},Int}()          # (v_q, lo, hi)   => var
+        pat = UInt64[]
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
Added the prefix-ladder order encoding trailed in the PR description (`3e71496`), since it panned out: every multi-version package gets biconditional ladder variables `L_k ≡ some version ≤ k chosen`, which (a) make at-most-one linear (`p@k ⇒ ¬L_(k-1)`, replacing the quadratic pairwise mutex) and (b) let interval conflict sides compile to `¬L_hi ∨ L_(lo-1)` directly, eliminating the interval trigger variables and their per-version definitions. The completion direction of the ladder is load-bearing: the positive `L_(lo-1)` literal is only correct because a chosen version in the interval forces the ladder prefix below it false.

| | pairwise (#42) | rectangles | + ladders |
|---|---|---|---|
| total clauses | 674k | 247k | **85k** |
| variables | 5.9k | 12.7k | 11.1k |
| SAT build | 60ms | 29ms | **14ms** |
| resolve end-to-end | ~355ms | ~279ms | **~265ms** |

Same verification as the base commit: byte-identical 200k-instance differential dump, identical registry solutions, full suite green. The tiny exhaustive testsets caught a ladder/package variable-numbering collision during development — three failing cases out of 373k — which is precisely the kind of encoding bug they exist to catch.

StefanKarpinski:
One more rung (`13f02e8`): dependencies were the largest remaining clause block (46.5k per-version binary clauses), and dep sets change rarely across versions — so they compress the same way, one clause per maximal run of depending versions using the ladder interval literals (a dep held by every version becomes a single package-level `¬p ∨ q`).

| | pairwise (#42) | rectangles | + ladders | + dep runs |
|---|---|---|---|---|
| total clauses | 674k | 247k | 85k | **42k** |
| SAT build | 60ms | 29ms | 14ms | **9ms** |
| resolve end-to-end | ~355ms | ~279ms | ~265ms | **~259ms** |

Same verification: byte-identical 200k dump, identical registry solutions, suite green. The encoding is now 16× smaller than where the day started, and the SAT build has gone from the dominant post-filter cost to a rounding error; what remains of resolve time is almost entirely the picosat solves themselves plus the (cacheable) provider and info build.

PR #44, opened 2026-07-29
Default each package's best version variable to true

StefanKarpinski:
Follow-on to #42/#43, from review discussion: rather than adding a batched rank-1 probe pass, bias the solver's phase defaults so models land at the optimum on their own. Unconstrained variables still default to false (keeping models junk-free), but each package's rank-1 version variable now defaults to true — whenever the solver must choose a version, it tries the best one first.

Benchmarked head-to-head on the DifferentialEquations closure (identical solutions in all variants):

| variant | descent | solves |
|---|---|---|
| per-package probe only (current) | 74.9ms | 55 |
| phase hints only | 68.1ms | 33 |
| **hints + existing probe (this PR)** | **45.5ms** | 35 |
| layer-batched probes, no hints | 92.6ms | 61 |
| hints + batched probes | 41.5ms | 26 |

The hints subsume what a batched-probe pass would buy: batching *without* hints is actively worse than the status quo (a failed all-layer probe is an expensive joint-UNSAT proof), and batching *with* hints saves only ~4ms more than this PR while adding a pass with that failure mode. One loop over packages at SAT construction wins instead.

Phase defaults cannot affect the answer (Theorem 1: the layered solution is solver-independent). Verified: byte-identical 200k-instance differential dump, identical registry solutions, full suite green. Resolve end-to-end ~240ms → ~205ms.

### DIFF
--- a/src/PicoSAT.jl
+++ b/src/PicoSAT.jl
@@ -18,6 +18,9 @@ add(p::Ptr{Cvoid}, v::Integer) =
     ccall((:picosat_add, lib), Cint, (Ptr{Cvoid}, Cint), p, v)
 assume(p::Ptr{Cvoid}, v::Integer) =
     ccall((:picosat_assume, lib), Cint, (Ptr{Cvoid}, Cint), p, v)
+set_default_phase_lit(p::Ptr{Cvoid}, lit::Integer, phase::Integer) =
+    ccall((:picosat_set_default_phase_lit, lib), Cvoid,
+          (Ptr{Cvoid}, Cint, Cint), p, lit, phase)
 set_global_default_phase(p::Ptr{Cvoid}, phase::Integer) =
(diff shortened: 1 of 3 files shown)


StefanKarpinski:
Per review discussion, added the layer-batched probe on top of the phase defaults (`4604af0`) — the "hints + batch" row from the variant table. Before optimizing a layer's packages sequentially, one solve assumes the best version of every package the current model doesn't already have at its best; joint feasibility witnesses each sequential optimization, so a successful probe pins the whole layer at once (and a failed one costs a single solve and tells us nothing — layers with one non-optimal package skip it, since the per-package probe covers them).

End-to-end interleaved measurement turned out better than the descent-level microbenchmark suggested: resolve ~271ms → ~210–240ms on the DifferentialEquations closure. Identical solutions (200k-instance dump byte-identical, suite green).

PR #45, opened 2026-07-29
Intern packages once at the top of the pkg_info build

StefanKarpinski:
The clarified "item 3" from the resolve-speed discussion: the filter passes already intern packages per call, but `pkg_info`'s three build phases still ran on `Dict{P,…}` lookups throughout — interaction pushes, offset tables, dependency unions, and every partner lookup during bit-setting. This interns once at entry and runs the whole build on integer ids (in name order, so id-sorted lists coincide with the name-sorted output), materializing package-keyed structures only at the end. The standalone validation pass folds into the id conversion with identical error messages.

Verification: serialized pkg info is **byte-identical** to main, both unfiltered and filtered, on the DifferentialEquations closure; 200k-instance differential dump identical; full test suite green.

Measured: unfiltered info build **115ms → 85ms** (best of 10, interleaved against a main worktree).

Independent of #44 (disjoint files).

### DIFF
--- a/src/PkgInfo.jl
+++ b/src/PkgInfo.jl
@@ -68,15 +68,33 @@ function pkg_info(
     reqs :: SetOrVec{P} = keys(data);
     filter :: Bool = true,
 ) where {P,V}
-    validate_pkg_data_consistency(data, reqs)
+    # intern packages as integer indices once, up front: the build phases
+    # below run entirely on vectors, and package names reappear only in
+    # the final PkgInfo structures. ids follow name order (packages are
+    # sorted), so id-sorted dependency and interaction lists coincide
+    # with the name-sorted ones the output requires
+    pkgs = sort!(collect(keys(data)))
+    N = length(pkgs)
+    ix = Dict{P,Int}(p => i for (i, p) in enumerate(pkgs))
+    datas = [data[p] for p in pkgs]
+    nver = Int[length(d.versions) for d in datas]
+
+    # required packages must exist
+    for r in reqs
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
With this merged, an overall before/after for the optimization campaign — pre-campaign main (`cf6eb67`, before #37) vs current main, interleaved best-of-runs on the same machine, answers verified identical on every instance. DifferentialEquations closure (26,373 raw versions, 91-package answer):

| phase | before | after | speedup | main contributor |
|---|---|---|---|---|
| provider (`pkg_data`, warm) | 401ms | 92ms | 4.3× | #40 dedup + entry grouping |
| info build (unfiltered) | 173ms | 85ms | 2.0× | #38 memoized masks, #45 interning |
| filter | 750ms | 52ms | **14.4×** | #38 kernels + batched dominance, #39 ordering |
| SAT build | 50ms | 8ms | 6.1× | #43 rectangles + ladders + dep runs |
| descent (derived) | ~228ms | ~29ms | ~8× | #42 probes + phases, #44 hints + layer batch |
| **resolve end-to-end** | **1201ms** | **174ms** | **6.9×** | |

Encoding: 709,772 → 42,155 clauses (16.8×). The filter also prunes slightly more than before (5,557 → 5,223 kept versions, from #39's iteration), and small instances (JSON, DataFrames) stay sub-millisecond with a ~5× provider improvement. Remaining profile: build 49%, filter 30%, descent 17%, SAT build 5% — the top two slices are requirement-independent and are what registry-state caching will remove during Pkg integration.

PR #57, opened 2026-07-31
Diagnostics: goal-query resolve, verified fixes, and holdback analysis on the production SAT instance

### LINKED ISSUE #6: Suggest what to remove for satisfiable requirements
I have a large environment with many packages. Is there an algorithm or heuristic that can suggest what to remove when I want to add a new thing?

The "unsatisfiable requirements" tree is too much for me to make sense of, sometimes.

### LINKED ISSUE #52: Diagnostics should explain steering surprises: why a resolve was held back to an older julia (or package) version
> **Build plan.** Rewritten 2026-09-09 as a handoff against `main` at `d59bcd2`. The comments below are history: the two 2026-08-17-and-earlier implementation sketches describe architectures that have since been replaced, and should be read for motivation only. This body is the current statement of what to build.

## The problem

A user compat bound on a non-upgradable stdlib steers the julia choice: a julia version is compatible with exactly the stdlib versions it bundles, so a stale `[compat] LinearAlgebra = "~1.10"` quietly pins the whole toolchain to julia 1.10. The resolution is *correct*. Nothing says why.

Measured instance in [the first comment below](https://github.com/StefanKarpinski/Resolver.jl/issues/52#issuecomment-5130394202): a project with `julia = "1.6"` and `LinearAlgebra = "~1.10"` resolves julia to 1.10.11 rather than 1.12.6, and the only signal the user gets is a `⊼` character (`bin/resolve.jl:774`).

**This is the last asymmetry in the diagnostics story.** An unsatisfiable resolve now gets a machine-verified proof per conflict, complete coverage of the cheapest fixes, and a per-action answer for every fix it does not offer. A *satisfiable but surprising* resolve gets one character.

Scope note: the surprise only exists when resolving **over** julia versions. Pkg pins julia to the running version by default, where the same stale bound produces an ordinary unsatisfiable resolve. But #62 names `holdbacks` as the mechanism for the version half of `pkg> why` — "why is X at 1.2 and not 1.9" — which is not julia-specific and is a headline Pkg feature.

## Prior art — a spec, **not** a patch

`archive/holdbacks-prototype` (commit `c9e8b0c`, "explain resolutions that succeeded and still surprised") is a complete prior implementation: `src/Holdback.jl` (361 lines), `test/holdback.jl` (230 lines), plus `bin/resolve.jl` wiring and docs. It settles a lot of design and is worth reading in full.

**It cannot be ported.** It targets `src/Diagnose.jl` and a `Fact`/`UserCompat`/`Bound` vocabulary that #105 replaced with clauses, and — the trap — **its tests assert wording that has since been retired**:

| prototype asserts | current idiom |
|---|---|
| `your compat restricts Stats to 10` | `your compat leaves Stats 10` |
| `julia 12 works with Stats only at 12` | directed verbs read off the consequent: `requires` / `constrains` / `leaves no version of` |

The symmetric "works with … only at" phrasing is exactly what #105 replaced, because a bound has no direction and which way it is said has to be *chosen* when printed (`src/Diagnostics.jl:2154`). Copy those assertions and you reintroduce two dead vocabularies. Take the prototype's **structure and edge cases**; re-derive every string.

Two sibling commits on the same branch are prior art for other roadmap items, not for this one: `1539af3` (goals) and `3993297` (the presentation toolkit and `--explain`). Same caveat applies to both.

## What to build

For a package `p` that resolved at class rank `r` when a better rank exists:

1. **Probe.** Assume `p` at a better rank on the solved instance. Satisfiable → nothing to explain. Unsatisfiable → continue.
2. **Blame.** Take a held-fixed MUS over the selector vocabulary — `sat_mus(satx, fixed, selectors)`, the same call `reason_core` makes at `src/Diagnostics.jl:1649`. The result names the responsible statements *as clauses*.
3. **Render.** Those clauses print through the existing `clause_phrase` / `range_phrase` path, so a holdback's lines read in the same idiom as a conflict's and no new sentence layer is needed.
4. **Verify the fix.** Re-resolve the relaxation on the universe already filtered for the query — licensed by #75's Theorem C, and the same route conflict fixes use. The witness is the optimal resolution under the fix, so it is as honest as an UNSAT fix's.

Everything in steps 1–4 exists and is tested. This is assembly, not new machinery.

## Decisions to make first

None of these is settled, and each shapes the code:

1. **Which packages get probed.** julia always? every requirement resolved below its best? everything? behind a flag? A probe is a solve, so this is a cost decision as well as a UX one. The prototype's answer was `max_probes = 8` with an explicit truncation line — see (4).
2. **Output shape.** Is a holdback `Conflict`-shaped — heading, `Line`s, a `Fix` with `Action`s — reusing `action_phrase` and printing inside the diagnosis? Or its own type printed beside it? The first buys the proof checker for free (see below); the second is easier to keep out of the unsatisfiable path.
3. **Entry point.** The prototype's was `holdbacks(info, prob, sol, [pkgs]; by, max_probes)`. Does `resolve` surface them on success, or does a caller ask separately? #62 needs an answer to write Pkg code against.
4. **Truncation honesty.** The prototype printed `(7 more packages resolved below their best version, not examined)`. That is the shape of footer **#108 just removed**, on #104's principle that the page owes only what can be acted on. There is now a precedent, and it may cut against the prototype: consider recording the count on the returned value and printing nothing.

## Verification — free, if the output is clause-shaped

`test/proof_check.jl` exports `lines_are_true`, `chain_is_a_proof`, `proof_problems`, `claimed_lines` and friends, and `lines_are_true(sat, prob, clauses)` is **generic over clauses**. If a holdback's lines are clauses over the same instance, they are proof-checked with no new checker: every line true of the universe the query left, and the claimed set contradictory.

That is a strong argument for decision (2) landing on "`Conflict`-shaped".

Harnesses to extend: `test/proof_check.jl` (the checker itself needs nothing), `test/chain_coverage.jl` (add holdback probes to the corpus loop), `test/diagnostics.jl` (report wording). The prototype's `test/holdback.jl` is the scenario list to port — `STEERING` and its siblings — with every `occursin` string rewritten.

## Where things are

- `src/Diagnostics.jl` — the diagnosis layer; `reason_core` (1729) and the selector plumbing (1793) are the models to follow.
- `src/UnsatCores.jl` — `sat_mus`, including the held-fixed form (135).
- `src/Clauses.jl` — the clause algebra and the sentence layer.
- `docs/src/theory/unsat-explanation.md` — the contract a report owes. A holdback is subject to it.
- `bin/resolve.jl` — the `⊼` marker (774) is the current signal; the prototype added `--explain` and automatic julia explanation here.
- **Docs target moved**: the prototype wrote `docs/src/diagnostics.md`, which no longer exists. Docs are now `index.md`, `api.md` and `theory/`.

## Definition of done

- Holdbacks are computed, blamed, rendered and fix-verified for the measured steering instance and for the prototype's scenario set.
- Every printed line passes `lines_are_true` against the instance the resolve ran on.
- The truncation decision from (4) is implemented and consistent with #108.
- `bin/resolve.jl` explains a held-back julia without being asked; the rest is behind a flag.
- `docs/src/api.md` documents whatever is public.

---

🤖 Rewritten as a handoff with [Claude Code](https://claude.com/claude-code).

StefanKarpinski:
Closes #6. Closes #52.

## What

When a resolve fails, `resolve` returns a `Diagnosis` explaining why and what would fix it. When it succeeds surprisingly — a package held below the version you'd expect — `holdbacks` names the constraint responsible. And both generalize to *goal queries*: ask the resolver for an outcome, and get either the optimal solution achieving it or the explanation of what it would take.

```julia
resolve(info, prob)                                  # solve, or explain why not
resolve(info, prob; with = "CUDA")                   # solve with CUDA, or what it would take
resolve(info, prob; without = "FillArrays")          # the FillArrays-free optimum, or why not
resolve(info, prob; with = "DataFrames" => v"1.8.2") # this version, or what it would take
   :: Union{Dict{P,V}, Diagnosis{P,V}, Nothing}      # nothing iff diagnose = false

issatisfiable(info, prob; with = ..., without = ...) # bare feasibility: one solve, no descent
holdbacks(info, prob, sol, [pkgs])                   # why is X at an old version?
```

`prob`'s constraints are *negotiable* — fixes may propose relaxing them; the goal kwargs are the *fixed ask* — never proposed as fixes. `diagnose=false` with a goal is a cheap feasibility probe; `issatisfiable` skips even the optimizing descent. A `Diagnosis` carries independent conflicts as minimal fact chains, **verified fixes** (each witness is the optimal resolve of the fixed problem — what you'd actually get), and ranked upstream suggestions:

```
Unsatisfiable — 1 conflict, 2 fixes:
  • you requested DataFrames at 1.7
  • DataFrames 1.4.0–1.7.1 requires PrettyTables at 2.1.0–2.4.0
  • you requested PrettyTables at 1
  • PrettyTables 1.3.1 works with DataFrames only at 0.11.7–0.21.8, 1.3.6
    (likewise for PrettyTables 0.1.0–0.11.1, 2.0.0–3.4.3, which your constraints already rule out)
Verified fixes:
  1. relax the version you requested for DataFrames → allows: PrettyTables 1.3.1, DataFrames 1.3.6
  2. relax the version you requested for PrettyTables → allows: DataFrames 1.7.1, PrettyTables 2.4.0
```

`bin/resolve.jl` prints the diagnosis on failure, explains julia automatically whenever it resolves below its best admissible version —

```
Note: julia is held back to 1.10.11 by your compat on LinearAlgebra; relaxing it allows julia 1.12.6.
```

— and answers `--explain=<pkg|pkg@ver>` on demand (below best → the holdback; at best or absent → what having/avoiding it costs).

## How

Diagnosis runs **on the same filtered, collapsed instance resolve used** — the selector-guarded user constraints (#49) turn into assumptions by popping one frame, and every diagnostic artifact is a satisfiability verdict under a relaxation. The licensing theory is new in `docs/src/theory/diagnostics.md`:

- **Theorem D1**: the filter preserves answers under every *column-closed* relaxation. The finer per-source version was conjectured and **disproved by the branch's own sweep** (Prop D1′ — minimal counterexample: a compat bound and a pin jointly forbidding both versions of a package; 1.4% of fine-grained relaxations answer wrongly). **Lemma R**: reachability alone is unconditionally relaxation-stable — redundancy elimination is the sole unstable pass.
- **Proposition G** (goal safety): better-version goals are safe on the per-resolve instance (reach keeps version prefixes); presence goals are not — a committed counterexample shows the filter calling a package "forced" that the raw problem avoids — so they run on a T1 closure sub-instance, exact by **Lemma D3**.
- ~1.9M swept relaxation verdicts back the theorems; goal queries are additionally checked against a brute-force oracle.

## Policy is conventions and post-processing, not knobs

The generation-time surface is `diagnose::Bool`. Everything else:

- Platform packages (julia) are modeled as *dependency edges*, never requirements — so "drop requirement julia" is unrepresentable, not suppressed.
- **Yanked versions are pre-filtered** from the universe: never suggested, by construction. `--allow-yanked[=<pkgs>]` opts back in; the report still explains, from provider metadata, when "the versions your compat admits are yanked."
- Prereleases stay a relaxable admission kind ("allow prereleases of Wine_jll" is a fix worth finding), suggested only when no same-base release supersedes them.
- **Every opinion in the report is a public function**: `report(io, d; max_upstream, demote_incidental, trim_witnesses, labels)`, `changes(sol₁, sol₂)`, `rank_upstream!`, `superseded`, the sentence layer, `Bound.incidental`. The default `show` is one client; a caller like Pkg owns its own presentation (guide: "Rendering your own reports"), including wording for `Pkg.add`-style restrictions via `labels` ("you requested DataFrames at 1.7").

## Cost

Automatic julia holdback: 4–18 ms (free when julia is at its best — no solve). Full UNSAT diagnosis on real registry conflicts: ~0.3–1.2 s end-to-end, with a deterministic work budget on pathological closures (the report then stops at requirement level and says so). Goal queries on a Makie-scale closure: `issatisfiable` 60–150 ms; full diagnosis 113–348 ms. Suite: ~1.9M assertions including the relaxation sweeps.

## Review history

Earlier iterations and review discussion: #30 (closed). Follow-on tracking: #59 (performance directions), #63 (output-quality ceilings), #64 (roadmap).

## Co-author

Implemented with [Claude Code](https://claude.com/claude-code): design discussed with, implementation by, and review by Claude, directed and reviewed by @StefanKarpinski, with review feedback from @KristofferC.

https://claude.ai/code/session_016p2sckwc5Gjjg5LG6gdEk7

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Resolver.jl separates *what* to optimize from *how* to solve, handing the solvin
     - a version's dependencies must be present, and
     - conflicting versions cannot both be chosen.
 - Before solving, the problem is shrunk by **reachability analysis** (keeping only versions that could appear in a solution) and **redundancy elimination** (dropping versions that are strictly dominated by others).
-- The solution is optimized **lexicographically, one package at a time**, against a caller-supplied priority order — the resolver returns the unique optimal resolution that order determines (`nothing` if the requirements are unsatisfiable), and it is **Pareto-optimal**: no valid resolution is strictly better.
+- The solution is optimized **lexicographically, one package at a time**, against a caller-supplied priority order — the resolver returns the unique optimal resolution that order determines, and it is **Pareto-optimal**: no valid resolution is strictly better.
 - Because the preference order is supplied per package, different strategies — newest, oldest/downgrade, prefer-installed, keep-current — can be mixed **within a single resolve**.
 
 The core resolver in [`src/`](src/) is generic: it knows nothing about Julia versions or registries. The [`bin/resolve.jl`](bin/resolve.jl) command-line tool adapts it to Julia's real registries and can emit a `Manifest.toml`.
@@ -38,7 +38,9 @@ The core resolver in [`src/`](src/) is generic: it knows nothing about Julia ver
 
 Experimental and research-stage. Resolver.jl is a standalone package rather than a part of `Pkg`, but the core resolver is complete, and `bin/resolve.jl` resolves real projects against the General registry today — it already backs [`julia-downgrade-compat`](https://github.com/julia-actions/julia-downgrade-compat), which uses it to compute minimum-version manifests in CI.
 
-A major in-progress goal is providing much better diagnostics when a manifest is **unresolvable** (unsatisfiable). The rich theory behind SAT problems helps here: rather than a bare failure, the resolver can compute
+An **unresolvable** manifest is explained rather than merely reported. The rich theory behind SAT problems is what makes that possible: instead of a bare failure, `resolve` returns a diagnosis computed from
 
 - **MUSes (minimal unsatisfiable subsets)** — smallest sets of requirements that already conflict on their own, explaining *why* resolution failed; and
-- **MCSes (minimal correction sets)** — smallest sets of requirements whose removal would make the manifest resolvable, pointing at *what to change* to fix it.
+- **MCSes (minimal correction sets)** — smallest sets of constraints whose removal would make the manifest resolvable, pointing at *what to change* to fix it.
+
+Both are computed on the very SAT instance the resolve failed on, which takes a theorem to justify — the filter that shrank that instance ran with all of those constraints in force. See the [diagnostics guide](docs/src/diagnostics.md) and the [theory](docs/src/theory/diagnostics.md) behind it, including the granularity below which the justification provably fails.
--- a/bin/Registries.jl
+++ b/bin/Registries.jl
@@ -1,7 +1,7 @@
 module Registries
 
 export registry_provider, package_info, bundled_versions,
-    prerelease_exclusion, yanked_exclusion, is_yanked, is_registered, is_bundled,
+    prerelease_exclusion, is_yanked, is_registered, is_bundled,
     UPGRADABLE_STDLIBS_UUIDS
 
 import Base: UUID
@@ -262,12 +262,14 @@ prerelease_exclusion(allow_pre::Dict{UUID,Bool}) =
 
 ## yanked versions
 #
-# Yankedness is the same shape of fact as prerelease-ness: a property of a version
-# that a query may or may not be willing to accept. Nothing offers the user a way
-# to accept one yet -- this kind is always in force -- but modeling it as a
-# constraint rather than a deletion is what keeps the package universe shared, and
-# is what lets a diagnostic eventually say "the only version that satisfies this
-# is yanked" instead of "no such version".
+# Unlike prerelease-ness, yankedness is *not* a query knob: a yanked version is
+# one the registry has withdrawn, and the resolver must never produce one and
+# never propose one. Both are true by construction if the version is simply not
+# in the universe, so the provider filters yanked versions out (and
+# `--allow-yanked` rebuilds without the filter, which is rare and cheap). What
+# is lost that way is only the *story* -- "no such version" where "the versions
+# your compat admits are yanked" is the real answer -- so the provider records
+# the versions it dropped, as explanation data rather than as a constraint.
 #
 # A version is offered when *some* registry entry has it un-yanked, so it counts
 # as yanked only when every entry that has it says so -- and a version that no
@@ -289,8 +291,12 @@ function is_yanked(
     return found
 end
 
-yanked_exclusion(packages::Dict{UUID,Vector{PkgEntry}}) =
-    :yanked => (uuid::UUID, v::VersionNumber) -> is_yanked(packages, uuid, v)
+# `allow_yanked` mirrors `allow_pre`: keyed by package uuid with the zero uuid
+# holding the default, so `--allow-yanked` and `--allow-yanked=Foo` both work.
+# Accepting a yanked version is the user's call to make, and the way to make it
+# is to build a universe that has them in it.
+allows_yanked(allow_yanked::Dict{UUID,Bool}, uuid::UUID) =
+    get(allow_yanked, uuid, get(allow_yanked, UUID(0), false))
 
 # Do the registries have this version, as opposed to it existing only because
 # some Julia bundles it? (the provider's `bundled_only`, from the outside.) Only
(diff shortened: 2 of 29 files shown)


StefanKarpinski:
Superseded. This branch was rebuilt from scratch on the architecture that landed since — #71 (interchangeability classes as the representation), #72 (the relaxation substrate) — and is shipping as a sequence of reviewable PRs instead: #73 (diagnose an unsatisfiable resolve) is open and green, with conflict-level stories and upstream suggestions, goals, holdbacks, and the presentation toolkit to follow.

Roughly half of what this PR contained is simply gone rather than ported. Class space made the per-source selector machinery, the firm/soft exclusion split, the D1/D1′ relaxation theory and its sweep, Proposition G's two universes, and the `mandatory`/`unrelaxable` policy knobs unnecessary; `Resolver.UnsatCores` (#69) replaced the hand-rolled MUS/MCS code. What survives — the fact vocabulary, verified fixes with witnesses from a real re-resolve, the rendering, holdbacks — survives in the new PRs.

Closing rather than retargeting: the diff no longer resembles what is being built.

Issue #70, opened 2026-08-05
Relaxation-stable filtering: make a query's filtered universe valid for every relaxation of it


StefanKarpinski:
Found while building the class-indexed universe (#58); recorded there rather than
fixed, because fixing it is a change to what licenses a redundancy deletion, not
an implementation detail. Nothing is wrong with any answer the resolver gives
today — this is about whether those deletions stay licensed once we start asking
what would happen if a user constraint were relaxed.

## The shape

A class competes at the rank of its best *admissible* member. So a compat bound
that forbids a class's best member does not remove the class; it moves it, and it
can move it behind a class it otherwise outranks. Concretely, with `P` at
versions `[v1, v2, v3]` where `v1` and `v3` are interchangeable:

- classes are `A = {v1, v3}` (rank 1, competing at `v1`) and `B = {v2}` (rank 2);
- the user forbids `v1`. `A` now competes at `v3`, so the rank order is `B`, `A`;
- in that order `B` dominates `A` — its registry row is a subset — so redundancy
  elimination deletes `A`;
- relax the compat and `A` competes at `v1` again, is once more the better class,
  and is the one an answer should return. It has been deleted.

Every counterexample the sweep found has this shape: a non-trivial class
permutation. Rate on random tiny grids: 14 / 11,200 queries (~0.13%).

## Why it is not a bug today

Domination is sound for the rank order it was computed in, and within one resolve
that order is fixed. The full oracle suite passes, including the brute-force
reference on raw `PkgData`, and the 48-scenario `bin/resolve.jl` comparison is
byte-identical to the pre-change resolver. What is at stake is only the stronger
property the diagnostics work wants: *nothing a relaxation can change may justify
a deletion.*

Note that this is not the same as the two things the class-indexed universe does
guarantee, which are tested in `test/class_space.jl`:

- a deactivated class is never deleted by redundancy, and never licenses a
  deletion (it is off both sides of the domination test);
- domination ignores activation, so a conflict against an emptied partner class
  still counts.

Those hold. The order is the remaining query-dependent input.

## Why the obvious fixes do not work

- **Rank classes by best member overall, ignoring admissibility.** Restores an
  order that constraints cannot move, but gives the wrong answer: a class whose
  best member is forbidden would outrank a class that can actually be selected at
  a better version.
- **Run redundancy in the registry order while the descent uses the query order.**
  Unsound outright — domination means "the better class will be selected
  instead", which has to be the order the descent selects in.
[...]

PR #73, opened 2026-08-05
Diagnose an unsatisfiable resolve

StefanKarpinski:
**Stacked on #72** (`relaxation-substrate`); GitHub retargets to `main` when that merges.

A failed resolve stops saying `nothing` and starts saying why. `resolve` returns a `Diagnosis`: the independent conflicts, each with a short chain of facts, and a menu of fixes that have each been checked by actually re-resolving the fixed problem. `diagnose = false` restores the bare `nothing` for callers that only want the verdict.

```
Unsatisfiable — 1 conflict, 3 fixes:

  • you require DataFrames
  • DataFrames: 0.11.7–0.22.7 are excluded by your compat
  • no version of PrettyTables is available: 0.1.0–3.4.5 are excluded by your compat

Verified fixes:
  1. relax your compat on PrettyTables
     → allows: DataFrames 1.8.2, PrettyTables 3.4.5
  2. relax your compat on DataFrames
     → allows: DataFrames 0.21.8
```

## How it works

Two things are assumable — "package p is installed" and "class c of p is forbidden" — and everything else about the problem is a clause the diagnosis never touches. On that footing the whole thing is four calls into `Resolver.UnsatCores` and one re-resolve:

- **Independent conflicts** are the disjoint minimal unsatisfiable subsets of the requirements.
- **A conflict's story** is a minimal unsatisfiable subset of that conflict's requirements plus the packages this query left with nothing installable, shrunk with the requirements attempted first so the surviving chain prefers the facts a user can act on.
- **The fix menu** is the minimal correction sets over the same candidates, enumerated with requirements *kept* first so the menu prefers relaxing a constraint over dropping a dependency.
- **A correction set becomes user actions** through #72's attribution: reactivating a class is not something a user does, so for each one we name the constraints that emptied it — cheapest member first.
- **Every fix is verified by re-resolving the modified problem from the artifact**, and the solution that resolve returns is the witness printed in the report. That is what "what you would get" means, and it is also what keeps #70's query-order caveat out of user-visible output: no version in a report is ever read off an instance filtered under constraints the fix relaxes.

Facts are per *package*, not per class. Per-class facts read as ten near-identical clauses for one package at registry scale, and they cannot be merged at render time without lying — a real case (`PrettyTables` pinned at 0.9.1) has an admitted version sitting between two excluded blocks. So a package's story is minimized over one literal standing for all of its emptied classes, introduced as a definitional extension in a scoped frame: the new variable occurs only negatively, so every model extends with it false and none is removed. The same granularity fixes the fix menu, where per-class enumeration produced *k* identical singleton corrections for one package and exhausted the limit on duplicates.

## Scope

Conflicts caused by registry compatibility between classes that are *not* empty are reported at requirement level — "A and B cannot be satisfied together" — without yet naming the bound responsible. Naming it, and suggesting the upstream release that would fix it, is the next PR. Goals, holdbacks and the presentation toolkit follow after that.

## Testing

`check_diagnosis` re-derives everything independently rather than trusting the implementation: each conflict's facts are asserted jointly unsatisfiable *and* individually load-bearing, each emptiness fact is checked against `class_exclusions` class by class, and — the central property — each fix's actions are applied to the `Problem` and re-resolved from the artifact, asserting the result equals the recorded witness. Swept over the tiny grids × constraint draws × admission kinds × two priority orders and two version orderings, with counters asserting the sweep actually produced multi-fix diagnoses and fixes that relax constraints rather than only dropping requirements. Plus: the instance is left exactly as it was found (verdict battery before and after, four shapes, each diagnosed twice), and a registry-scale case end to end through `bin/resolve.jl`.

## Co-author

Implemented with [Claude Code](https://claude.com/claude-code): implementation and review by Claude, directed and reviewed by @StefanKarpinski.

https://claude.ai/code/session_016p2sckwc5Gjjg5LG6gdEk7

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Resolver.jl separates *what* to optimize from *how* to solve, handing the solvin
     - a version's dependencies must be present, and
     - conflicting versions cannot both be chosen.
 - Before solving, the problem is shrunk by **reachability analysis** (keeping only versions that could appear in a solution) and **redundancy elimination** (dropping versions that are strictly dominated by others).
-- The solution is optimized **lexicographically, one package at a time**, against a caller-supplied priority order — the resolver returns the unique optimal resolution that order determines (`nothing` if the requirements are unsatisfiable), and it is **Pareto-optimal**: no valid resolution is strictly better.
+- The solution is optimized **lexicographically, one package at a time**, against a caller-supplied priority order. If the problem is satisfiable, the resolver returns the unique Pareto-optimal resolution that version order and package priority determine. If it is unsatisfiable, it returns a **diagnosis**: which requirements and user constraints are in conflict, why, and a menu of verified fixes.
 - Because the preference order is supplied per package, different strategies — newest, oldest/downgrade, prefer-installed, keep-current — can be mixed **within a single resolve**.
 
 The core resolver in [`src/`](src/) is generic: it knows nothing about Julia versions or registries. The [`bin/resolve.jl`](bin/resolve.jl) command-line tool adapts it to Julia's real registries and can emit a `Manifest.toml`.
--- a/src/Diagnostics.jl
+++ b/src/Diagnostics.jl
@@ -1,0 +1,618 @@
+"""
+    Resolver.Diagnostics
+
+Why a resolve failed and what would make it succeed. `diagnose` puts further
+questions to the universe a failed resolve was run against and the instance it
+failed on, and answers with a [`Diagnosis`](@ref): the independent
+[`Conflict`](@ref Resolver.Diagnostics.Conflict)s — sets of requirements that
+cannot hold together, each with a chain of
+[`Fact`](@ref Resolver.Diagnostics.Fact)s about the packages that says why —
+and a menu of [`Fix`](@ref Resolver.Diagnostics.Fix)es, each a set of
+[`Action`](@ref Resolver.Diagnostics.Action)s the user could carry out together
+with the solution that carrying them out yields.
+
+A `Diagnosis` is plain data, so a caller can render its own report from one
+without asking the resolver anything further; `show(io, MIME("text/plain"), d)`
+and [`action_phrase`](@ref Resolver.Diagnostics.action_phrase) are the report
+this module writes.
+
+These names are this module's API. `Resolver` exports `Diagnosis`, since it is
+what `resolve` answers with when the requirements cannot be satisfied, and
+re-exports none of the others.
+"""
+module Diagnostics
+
+export Diagnosis, Conflict, Fix, Action, Fact, Requirement, Availability,
+    unavailable, action_phrase, diagnose
+
+using ..Resolver: SAT, Problem, Universe, relax, resolve, installed_lit,
+    with_temp_clauses, with_classes_relaxed, sat_new_variable, sat_add_var,
+    sat_add, exclusion_kinds, class_exclusions
+using ..UnsatCores: sat_mus, sat_disjoint_muses, sat_mcses
+
+# What an unsatisfiable resolve says, and how it is found.
+#
+# A diagnosis answers two questions about a query the resolver could not
+# satisfy: which requirements cannot hold together, and what the user could
+# change so that they can. Both are questions about *relaxations* of the
+# problem — the same problem with some requirements dropped and some of the
+# classes this query emptied made choosable again — and both are put to the
+# universe the resolve was run against and the instance it failed on, before
+# they are discarded.
+#
+# Two things can be relaxed, and nothing else:
+#
+#   * a requirement, which the instance states as `installed_lit(sat, p)`;
+#   * a class this query emptied, which it states as `forbidden_lit(sat, p, c)`.
+#
+# Everything else is a clause the diagnosis never touches: dependency edges,
+# registry conflicts, the structural at-most-one, and the classes that are not
+# empty. Every query below is therefore a solve under a subset of one fixed set
+# of assumptions — which is what `UnsatCores` answers, and what the licensing
+# argument on the manual's Diagnostics page covers.
+#
+# Candidate order is the one knob those answers take, and both uses here put
+# the requirements first, for reasons that happen to agree: a repair is grown
+# by keeping candidates in order, so requirements are kept and the menu prefers
+# relaxing a constraint to dropping a dependency; a conflict is shrunk by
+# deleting candidates in order, so a requirement the story does not need drops
+# out of it.
+#
+# Both are asked a *package* at a time, over one assumable literal per package
+# the query left short (`with_emptied_packages`). That is the granularity of
+# both answers: a report says what has become of a package's versions, and a
+# fix says which of a package's constraints to relax, neither of which is a
+# sentence about one class. It is also what keeps the menu short — a package's
+# classes usually go the same way, and one repair per class of each of them
+# would be the same handful of instructions over and over.
+#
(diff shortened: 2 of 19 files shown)


src/Resolve.jl
```diff
@@ -91,15 +91,22 @@ function resolve_core(
 end
 
 """
-    resolve(data, prob::Problem; by = identity) -> Union{Dict{P,V}, Nothing}
-    resolve(data, reqs; by = identity, compat = ..., pins = ...)
+    resolve(data, prob::Problem; by = identity, diagnose = true)
+        -> Union{Dict{P,V}, Diagnosis{P,V}, Nothing}
+    resolve(data, reqs; by = identity, compat = ..., pins = ..., diagnose = true)
```
StefanKarpinski:
```suggestion
    resolve(data, reqs; compat = ..., pins = ..., by = identity, diagnose = true)
```
Better to put the (reqs, compat, pins) arguments together since they compose in a Problem instance. Similarly elsewhere.

src/Diagnose.jl
```diff
@@ -0,0 +1,572 @@
...
+unavailable(f::Emptiness) = all(!isempty, f.excluded)
+
+"""
+    Resolver.Action{P}(kind, pkg)
+
+One thing the user could change about `pkg`:
+
+  * `:drop` — stop requiring it;
+  * `:compat` — relax the allowed-version set constraining it;
+  * `:pin` — unpin it;
```
StefanKarpinski: This should really be called `:unpin` rather than `:pin`.

src/Diagnose.jl
```diff
@@ -0,0 +1,572 @@
...
+version back.
+"""
+struct Emptiness{P,V} <: Fact
+    pkg      :: P
+    members  :: Vector{V}
+    excluded :: Vector{Vector{Symbol}}
+end
+
+"""
+    Resolver.unavailable(fact::Emptiness) :: Bool
```
StefanKarpinski: Should this not be called `Resolver.excluded`? Also, either way, `Resolver.unavailable` or `Resolver.excluded` seem like arguments for putting the `Diagnosis` machinery in its own namespace. Having `Diagnosis.excluded` / `Diagnosis.unavailable` seems much more specific and clear than `Resolver.excluded` / `Resolver.unavailable` which feels like it might be involved in the resolution algorithm rather than diagnosis.

src/Diagnostics.jl
```diff
@@ -0,0 +1,572 @@
...
+[`Fix`](@ref Resolver.Fix)es, each of which repairs the whole problem.
+
+`show(io, MIME("text/plain"), d)` prints the report; the fields are plain data,
+so a caller that wants a different report can build one from them without
+asking the resolver anything further.
+
+Pass `diagnose = false` to [`resolve`](@ref) for the bare `nothing` when only
+the verdict is wanted.
+"""
+struct Diagnosis{P,V}
```
StefanKarpinski, line 229: Since the type name `Diagnosis` would clash with a module named `Diagnosis` (modules have an automatic self-binding), the module could be called `Diagnostics`.

src/Diagnose.jl
```diff
@@ -0,0 +1,572 @@
...
+# named constraint source taken off the package it names. An admission kind is
+# stated about versions rather than packages, so relaxing one means excepting
+# that package from it rather than dropping it.
+function relaxed_problem(
+    prob    :: Problem{P},
+    actions :: Vector{Action{P}},
+) where {P}
+    drop   = Set{P}(a.pkg for a in actions if a.kind === :drop)
+    uncomp = Set{P}(a.pkg for a in actions if a.kind === :compat)
+    unpin  = Set{P}(a.pkg for a in actions if a.kind === :pin)
```
StefanKarpinski: See we have `unpin` vs `:pin` here? They should match.

src/Diagnose.jl
```diff
@@ -0,0 +1,572 @@
...
+) where {P,V}
+    vm = VarMap(sat)
+    # the same package required twice is one thing to assume and one to drop
+    reqs = unique(prob.reqs)
+    # a requirement the universe holds no installable version of is not
+    # something the instance can be asked about: it has no variable, and no
+    # constraint of this query is what took it away, since a constraint empties
+    # classes rather than deleting them. It is a conflict of its own, and every
+    # fix has to drop it
+    gone = P[p for p in reqs if !haskey(sat.vars, p)]
```
StefanKarpinski: Now that VarMap is a thing, should we be querying that object instead?

README.md
```diff
@@ -29,7 +29,7 @@ Resolver.jl separates *what* to optimize from *how* to solve, handing the solvin
     - a version's dependencies must be present, and
     - conflicting versions cannot both be chosen.
 - Before solving, the problem is shrunk by **reachability analysis** (keeping only versions that could appear in a solution) and **redundancy elimination** (dropping versions that are strictly dominated by others).
-- The solution is optimized **lexicographically, one package at a time**, against a caller-supplied priority order — the resolver returns the unique optimal resolution that order determines (`nothing` if the requirements are unsatisfiable), and it is **Pareto-optimal**: no valid resolution is strictly better.
+- The solution is optimized **lexicographically, one package at a time**, against a caller-supplied priority order — the resolver returns the unique optimal resolution that order determines (a **diagnosis** if the requirements are unsatisfiable: which of them cannot hold together, why, and a menu of verified fixes), and it is **Pareto-optimal**: no valid resolution is strictly better.
```
StefanKarpinski:
```suggestion
- The solution is optimized **lexicographically, one package at a time**, against a caller-supplied priority order. If the problem is satisfiable, the resolver returns the unique Pareto-optimal resolution that version order and package priority determine. If it is unsatisfiable, it returns a **diagnosis**: which requirements and user constraints are in conflict, why, and a menu of verified fixes.
```

src/Diagnostics.jl
```diff
@@ -0,0 +1,622 @@
...
+A `Diagnosis` is plain data, so a caller can render its own report from one
+without asking the resolver anything further; `show(io, MIME("text/plain"), d)`
+and [`action_phrase`](@ref Resolver.Diagnostics.action_phrase) are the report
+this module writes.
+
+These names are this module's API. `Resolver` exports `Diagnosis`, since it is
+what `resolve` answers with when the requirements cannot be satisfied, and
+re-exports none of the others.
+"""
+module Diagnostics
```
StefanKarpinski, line 23: The module name and the file name should match.

src/Diagnose.jl
```diff
@@ -0,0 +1,622 @@
...
+The fact that the query requires `pkg`.
+"""
+struct Requirement{P} <: Fact
+    pkg :: P
+end
+
+"""
+    Resolver.Diagnostics.Emptiness{P,V}(pkg, members, excluded)
+
+The fact that the query's own constraints take versions of `pkg` away from the
```
StefanKarpinski: What does this type actually represent? This doesn't really say...

src/Diagnose.jl
```diff
@@ -0,0 +1,622 @@
...
+
+"""
+    Resolver.Diagnostics.Action{P}(kind, pkg)
+
+One thing the user could change about `pkg`:
+
+  * `:drop` — stop requiring it;
+  * `:compat` — relax the allowed-version set constraining it;
+  * `:unpin` — unpin it;
+  * anything else — admit the versions that admission kind forbids, for this
```
StefanKarpinski: Aren't we calling these "exclusions"?

PR #76, opened 2026-08-07
Answer a relaxation on the universe already filtered for the query

StefanKarpinski:
Follows #75, and is what makes its theorem usable.

#75 proved that a universe filtered for query `Q` contains everything an answer
to any relaxation `R ≤ Q` needs (Theorem C). Nothing in `src/` could act on
that: every public `resolve` routes through `prepare_pkg_info`, which ranks
*and* filters, so there was no way to say "take this universe and answer a
relaxation of the query it was filtered for". The only implementation was a
ten-line helper in `test/relaxation_stable.jl`, written to check the theorem
rather than to be used.

`resolve(sat, rp::RelaxedProblem)` is that call. It returns what
`resolve(info, rp.prob)` returns, running no pass of the filter and building no
second instance, and leaves the instance as it found it so one universe answers
as many relaxations as it is asked. The consumer is diagnostics: verifying a
candidate fix becomes a re-rank and a descent instead of a re-prepare.

Three commits: a one-line fix to a flaky coverage guard this PR's tests would
otherwise inherit, then the frame-depth mechanism, then the relaxation work
that uses it.

## The value, not the arguments

```julia
sources(prob)                                            # the query's constraint sources
relax(univ, Q, drop_reqs, drop_sources; order = nothing) # -> RelaxedProblem
resolve(sat,  rp::RelaxedProblem; by = identity)
resolve(univ, rp::RelaxedProblem; by = identity)
```

An earlier revision took `(sat, Q, R; order)`, which made two things the
caller had to get right and could not be helped with: that `R` was a relaxation
of `Q`, and that `order` matched what the universe was ranked in. Both are gone
rather than documented or checked. `relax` *derives* the relaxed problem by
withdrawing demands instead of accepting one, so being a relaxation is a
consequence of how the value was made rather than a claim about an argument;
and it settles there what every solve would otherwise recompute — the
representatives and the class ranking — so `order` is consumed at construction
and never reaches a solve.

`relax(univ, Q, Q.reqs, sources(Q))` is the bottom of the lattice;
`relax(univ, Q)` is `Q` itself; a diagnostics fix is a subset of `sources(Q)`
plus a list of requirements, which is the shape a fix already has.

That also stops repeated work: `version_permutations` was recomputed on every
solve. For the canonical ordering it returns immediately, but for a
non-canonical one it walks every package and sorts the ones out of order —
0.17–0.82 ms per solve on these roots, up to a quarter of the cheapest solves.
It is now paid once per relaxation.

## Re-ranking invalidates no clause

Every clause states which classes exist, depend and conflict — including the
prefix-ladder ones, whose "some class at layout position ≤ k" is a structural
fact used for at-most-one and the interval conflict encoding. A relaxation
changes which class we *prefer*, not which classes there are.

So the universe is not relaid out. The new ranking is threaded through the
descent instead, as the two questions the five "better than the current class"
sites actually ask: which class to probe first, and which classes an
improvement step may demand. `LayoutOrder` answers them with `1` and `1:i-1`,
so the ordinary descent is the code it was — **descent allocations are
byte-identical before and after** (12,976 / 20,056 / 62,464 / 84,928 for JSON,
DataFrames, DifferentialEquations, Makie), and the timing differences are
inside the run-to-run noise.

A package's ranking is just a vector the descent indexes: `ord_p[1]` is the
best class, and the classes ranking above the one it holds are the ranks the
scan passes on the way to it. For a relaxation that reorders nothing — which is
every relaxation of these four registry roots, and ~93% even under a generator
built to force reordering — that vector is `Base.OneTo(n)`.

Where something did move it is a `SparsePerm <: AbstractVector{Int}`: a
permutation stored as the pairs `(i, p[i])` for the positions it actually
displaces. `class_ranking` builds it while it ranks — sorting into a scratch
buffer reused across the whole pass, then walking it and pushing what moved — so
a package that reorders nothing allocates nothing and one that swaps two classes
carries two pairs. Nothing dense is retained. Being an `AbstractVector` is what
lets it flow unchanged through `copy_ranked!`'s `members[perm]`, `reps[p][perm]`,
`invperm` and `relaid_conflicts`, and `relax` keeps exactly what `class_ranking`
built, with no conversion.

That is the whole abstraction:

```julia
ranking(::Nothing, p, m::Int) = Base.OneTo(m)
ranking(ord::Dict{P,SparsePerm}, p::P, m::Int) where {P} = get(ord, p, Base.OneTo(m))
```

There is no inverse of it in the descent. "Is `p` at its best" is
`sol[p] == ord_p[1]`, and the better-set is a forward scan that stops at the
class `p` holds — which is guaranteed to be in `ord_p`, and which emits exactly
the literals the caller was going to emit anyway. `RelaxedProblem` is
parameterized on the ordering's type, so the one dynamic dispatch lands at
`resolve(sat, rp)` and everything behind that barrier is specialized.

## `reps` moves under a relaxation

`sat.reps[p][c]` is the member class `c` stands for *under `Q`* — a class whose
best member `Q` forbade stands at a better one under `R`. So `resolve_core`
hands back classes and the caller names them with `R`'s own representatives;
`sat.reps` is never mutated. `class_ranking` returns the new naming map and the
new ordering in one call, because they are one computation.

Deactivation is the other half of what `reps` means, and that is structural:
the reuse path swaps `Q`'s forbidding frame for `R`'s using #72's substrate,
scoped so an exception cannot strand the instance in `R`'s state.
`with_classes_relaxed` is now literally the empty end of that operation.

Frame depth is a field on `SAT` maintained by `sat_push`/`sat_pop`, and the
`with_*` helpers assert balance on entry *and* exit, so an unbalanced push
inside a body is caught at the call that did it. (`picosat_context` cannot
serve: it returns the context literal, not a level, and picosat recycles them.)

## The one precondition left

`reps` and the class ranking are indexed by the layout of the universe `relax`
was given, so a `RelaxedProblem` is stale if that universe is filtered again
afterwards. Nothing does that — `drop_unmarked!` is reachable only from
`prepare_pkg_info`, which builds a universe rather than shrinking one already
in use — so it gets one sentence in the docstring and no runtime machinery. A
`frozen` flag on `Universe` was considered and declined: it would prevent a
sequence nobody performs.

An earlier revision checked at runtime that `R` was a relaxation of `Q`. That
was 24% of the change and a parameter existing for no other purpose, to guard
an internal calling convention — where `resolve_prepared(univ, prob)`,
immediately above, carries the same class of precondition and documents it in a
one-line comment. What verifies it now is a test asserting that *`relax` itself*
produces relaxations: `reqs_R ⊆ reqs_Q` and `adm_Q ⊆ adm_R` pointwise, Lemma 1's
hypothesis, once per relaxation built.

## Numbers

Reuse against `resolve(info, R)`, min of 7, five relaxations each:

| root | per relaxation | batch of five |
|---|---|---|
| JSON | 2.3× – 17.4× | 34.8 → 4.8 ms (7.3×) |
| DataFrames | 2.1× – 7.1× | 49.5 → 14.3 ms (3.5×) |
| DifferentialEquations | 3.6× – 13.4× | 754.7 → 103.7 ms (7.3×) |
| Makie | 2.3× – 9.2× | 439.1 → 96.7 ms (4.5×) |

The batch figure includes `relax` itself (0.01–1.30 ms); the per-relaxation
range is the solve alone.

Every reuse answer is asserted equal to the re-prepared one before timing.

## Tests

The Theorem-C sweep drives `resolve_relaxed` itself, so what proved the theorem
exercises the production path rather than a parallel implementation of it. Its
primary assertion — that reuse answers `R` exactly as a fresh resolve does — is
what catches this whole class of bug, but only on queries that actually
*reorder* classes, and it was barely reaching those: 0.5–1.3% of relaxations.

The reason turned out to be structural. A class of one member competes at that
member whatever the query says, so its key is identical under every relaxation
— it cannot be reordered even in principle. The sweep's data was producing 19
multi-member classes out of 746. Sparse dependency and compat patterns (the
trick `test/classes.jl` already uses, for the same reason) take that to 105 of
637, and a second query regime forbids the best member of every multi-member
class on purpose. Reordering now fires on **5.6–7.4% of relaxations overall and
11.4–15.2% within the demoting regime**; the assertion's threshold sits ~4σ
below the observed band rather than inside it, so it does not flake.

Alongside: a relaxation that revives a class `Q` emptied and must name the
version `R` admits; a hand-built reordering case; one instance answering many
relaxations in fresh orders with no residue; frame-balance assertions catching
both a wrongly-nested call and an unbalanced body; and `ClassRanking` against
dense `randperm` permutations.

Full suite green, docs build clean, and the package precompiles — that last
one is checked explicitly now, because an earlier revision of this branch had
an outer constructor silently overwriting the one Julia generates, which the
full suite tolerated and only the docs build surfaced.

## Disclosure

While reviewing this branch I saw `bin/ tooling regression tests` fail once, on
an early commit, in a way that has not reproduced: every subsequent full-suite
run passes, on both sides of a rebase, and a full-suite control on `main`
passes. The branch touches no `bin/` file. I believe it is a flake — those
tests hit the live registry — but I did not explain it, and would rather say so
than let "suite green" imply it was never seen. #77 is why the next occurrence
will come with the child's output attached.

### DIFF
--- a/src/Resolve.jl
+++ b/src/Resolve.jl
@@ -5,8 +5,9 @@ function resolve_core(
     sat  :: SAT{P},
     reqs :: SetOrVec{P} = keys(sat.info);
     by   :: Function = identity, # priority ordering
+    ord  :: O = nothing, # class ranking (the universe's own layout by default)
     restore :: Bool = true, # restore the SAT instance's state before returning
-) where {P}
+) where {P, O}
     # a requirement the instance doesn't know has no installable version —
     # the filter drops such packages — so it cannot be satisfied
     all(p -> haskey(sat.info, p), reqs) || return nothing
@@ -31,15 +32,17 @@ function resolve_core(
         # below, so the layered answer is unchanged. a failed probe
         # tells us nothing and the sequential path proceeds as usual
         # (skipped for a single package, whose own probe covers it)
-        todo = [p for p in layer if sol[p] > 1]
+        todo = [p for p in layer
+                if sol[p] != @inbounds ranking(ord, p, nclasses(sat.info[p]))[1]]
         if length(todo) > 1
             for p in todo
-                sat_assume(sat, p, 1)
+                sat_assume(sat, p,
+                    @inbounds ranking(ord, p, nclasses(sat.info[p]))[1])
             end
             is_satisfiable(sat) && extract_solution!(sat, sol)
         end
         for p in layer
-            optimize_version!(sat, sol, p)
+            optimize_version!(sat, sol, p, ord)
             # fix optimized class
             sat_add(sat, p, sol[p])
             sat_add(sat)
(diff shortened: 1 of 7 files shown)


src/SAT.jl
```diff
@@ -318,79 +318,100 @@ SAT(info :: Dict{P,PkgInfo{P,V}}) where {P,V} = SAT(Universe(info))
...
+        finally
+            sat_pop_if(sat, !isempty(deact))
+        end
     finally
         push_deactivations!(sat)
         @assert sat.depth == depth balance_error(sat, depth)
     end
 end
 
+sat_pop_if(sat::SAT, yes::Bool) = yes ? sat_pop(sat) : sat
```
StefanKarpinski: This is kind of a dubious function. Seems just as clear to inline this at every usage.

src/Resolve.jl
```diff
@@ -162,6 +165,148 @@ function resolve_prepared(
...
+the one `univ` was ranked in: it is what "better" means, and the universe and the
+relaxation have to mean the same thing by it. It is consumed here, so the
+permutations it induces are computed once however often the result is resolved.
+"""
+function relax(
+    univ :: Universe{P,V},
+    Q    :: Problem{P},
+    drop_reqs = (),
+    drop_sources = ();
+    order = nothing, # version ordering
```
StefanKarpinski: Shouldn't these be typed?

src/SAT.jl
```diff
@@ -318,79 +318,98 @@ SAT(info :: Dict{P,PkgInfo{P,V}}) where {P,V} = SAT(Universe(info))
...
-# Inside, activation is entirely a matter of assumption: assuming
-# `forbidden_lit(sat, p, c)` forbids class `c` of `p` for one solve, and
-# assuming all of `sat.deact` states exactly what the frame stated, so any
-# subset of the query's deactivations can be imposed one solve at a time with no
-# clause added and none rewritten. Restoring re-asserts the same unit clauses,
-# because popping a frame discards the clauses asserted in it.
+# The query's own frame, re-asserted. Which classes it forbids and where each
+# package's phase hint points are properties of the query, true whether or not
+# the frame is in force, so `deactivate_classes!` settles them once, around this.
+push_deactivations!(sat::SAT) = push_forbidden!(sat, sat.deact)
```
StefanKarpinski: This feels like kind of an unnecessary method and obscures the fact that `push_forbidden!(sat, deact)` and `push_deactivations!(sat)` do fundamentally similar things.

src/SAT.jl
```diff
@@ -496,31 +515,91 @@ function with_temp_clauses(body::Function, sat::SAT)
...
+# `(to, to_ord)` does. A phase is a solver knob rather than a clause, so no
+# `sat_pop` takes one back down: an instance that answers one relaxation after
+# another has to have each one's hints taken down explicitly before the next
+# one's go up, or the first relaxation's guesses leak into the second's solves.
+# Settled once per relaxation and not touched again — once the descent pins a
+# package it adds a unit clause, and the hint stops mattering.
+function rehint_classes!(
+    sat      :: SAT{P},
+    from     :: Dict{P,Vector{Int}},
+    from_ord,
```
StefanKarpinski: Shouldn't this be typed?

src/SAT.jl
```diff
@@ -496,31 +515,91 @@ function with_temp_clauses(body::Function, sat::SAT)
...
+    to       :: Dict{P,Vector{Int}},
+    to_ord,
```
StefanKarpinski: This should also be typed.

PR #78, opened 2026-08-11
A problem's constraints are one namespace, keyed by kind

StefanKarpinski:
A `Problem`'s constraints were three shapes — `compat` a package-keyed dict of
allowed-version sets, `pins` a package-keyed dict of versions, `excludes` a
list of `kind => predicate` pairs — but every consumer already treated them as
one namespace keyed by symbol. `exclusion_sources` said so plainly:

```julia
s === nothing || v ∈ s || push!(srcs, :compat)
w === nothing || v == w || push!(srcs, :pin)
for (kind, forbids) in prob.excludes
    forbids(p, v)::Bool && push!(srcs, kind)
end
```

One vocabulary at the point of use, three things at the point of storage, and
every consumer paying for the difference with a hard-coded three-way branch:
`is_excluded`, `exclusion_sources`, `exclusion_candidates`, `is_constrained`,
`relax`, `sources`. Code downstream then invented translation layers and
reserved names to bridge the seam.

Now there is one map, `prob.constraints`, from **kind** to constraint.
`:compat` and `:pin` are well-known kinds; anything else is its own. A
constraint answers three questions and nothing else:

```julia
excludes(c, p, v) :: Bool                  # does this forbid v of p?
named(c) :: Union{Nothing, Set{P}}         # the packages it speaks about
relax(c, pkgs) :: Union{C, Nothing}     # it, no longer applying to pkgs
```

`is_excluded` is a loop. `exclusion_kinds` is a loop. `relax(prob, …)` is a
loop.

## `nothing` means one thing

`named` returns `nothing` for a constraint that speaks about *every* package: a
predicate cannot enumerate them, while a compat bound or a pin names its own.
So relaxing a kind outright is relaxing it for `named(c)`, and

```julia
relax(c, named(c)) === nothing
```

holds for every shape. "Withdrawn outright" needs no convention and no
sentinel, which is what makes the map safe to consult with `haskey`.

## Types

```julia
struct Problem{P, C<:AbstractDict{Symbol}}
    reqs :: Vector{P}
    constraints :: C
end

const Constraint{P,V,S} = Union{Compat{P,S}, Pins{P,V}, Predicate{P}}
```

`C` is a parameter, so it is always concrete — the shared empty singleton for
an unconstrained problem, a `Dict` otherwise — and the value type is a closed
union that `is_excluded` splits on rather than dispatching through.

This is *better* than before, not merely preserved: `compat` and `pins` were
`AbstractDict` fields, so every `get` on them was a dynamic call; `Compat` and
`Pins` now hold concrete `Dict`s, which costs nothing because an unconstrained
problem carries no such constraint at all. `code_typed` on `is_excluded` shows
an `isa` chain with the compat lookup inlined to `memoryrefget`, and it
allocates nothing.

**Correction.** An earlier version of this description claimed prepare
allocations drop 4.5–10%. That figure came from a harness I did not run myself,
and it does not reproduce: measuring `main`, this branch and the alternative in
#79 with one script, prepare allocations are identical to within 0.03% on a
constrained query and byte-identical on an unconstrained one, with times inside
this machine's noise. There is no measured performance difference either way.

## Two behaviour changes, both intended and tested

- **Every kind relaxes per package**, `:compat` and `:pin` included. Before,
  a caller could release a whole compat entry or a whole predicate, but not a
  predicate *for one package* — which is the shape a repair takes ("allow
  prereleases of X"). The new testset runs the same six-case matrix over
  `:compat`, `:pin` and a predicate.
- **Two constraints may no longer share a kind.** `Problem(reqs; compat = …,
  excludes = [:compat => f])` throws. That is what makes the map a namespace
  rather than a bag; nothing in the repo did it.

## The constructor says it too

```julia
Problem(reqs; compat = Dict(p => spec), pin = Dict(q => v), pre = (p, v) -> ispre(v))
```

Every keyword is a kind, and the keyword's name *is* the kind's name. `compat`
and `pin` want a package-keyed dictionary; anything else wants a predicate. A
custom kind is spelled the same way as a built-in one instead of being
relegated to a list of pairs.

That deletes the duplicate-kind check outright: a repeated keyword is a syntax
error, raised at lowering, before the call exists.

```
julia> Problem([:A]; compat = Dict(), compat = Dict())
ERROR: syntax: keyword argument "compat" repeated in call to "Problem"
```

Shape checking moved into the constructor, which is what keeps a typo an error
rather than a silent constraint:

```
julia> Problem([:A]; compt = Dict(:B => [:v1]))
ERROR: ArgumentError: constraint kind :compt wants a predicate, `(p, v) -> Bool`
true for the versions it forbids; got a Dict{Symbol, Vector{Symbol}}. `compat`
and `pin` are the only keywords that take a dictionary
```

`resolve` and `issatisfiable` **do not** slurp, and lose their constraint
keywords: their keyword space already holds `by`, `order` and `diagnose`, and
`order` takes a callable — so a typo'd `oder = my_order` would have slurped into
a predicate kind, passed the callable check, and silently applied a nonsense
constraint. The requirements-only methods stay; a constrained resolve goes
through a `Problem`, and `resolve(data, reqs; compat = …)` is now a
`MethodError`.

## Renames

`sources` → `relaxations`, `exclusion_sources` → `exclusion_kinds` (now sorted,
since a `Dict` has no order), the `pins` keyword → `pin` (the keyword's name is
the kind's name now, and `:pin` is how it reads at every point of use), and
"source" retired in favour of "kind". `NoKinds` goes with `excludes`.

## It is not smaller

`Problem.jl` grows about 135 lines for the three constraint types and the shape
checking, against 26 saved in `Resolve.jl`; the test churn is the `pins` → `pin`
and `excludes` → keyword rewrites across ~139 call sites. What it buys is one
vocabulary instead of three, which is what the diagnosis work needs — it keys
repairs by these same symbols, and against the old storage it had to invent a
second set of names and translate between them.

### DIFF
--- a/bin/Registries.jl
+++ b/bin/Registries.jl
@@ -248,14 +248,15 @@ is_bundled(uuid::UUID, v::VersionNumber) =
 #
 # Whether a query will accept prerelease versions is a *query* fact, not a
 # registry one: the versions exist either way. So the provider offers them all
-# and `--allow-pre` reaches the resolver as one of the `Problem`'s exclusion
-# kinds, forbidding the prereleases of the packages it was not given for -- which
-# is what deleting them used to accomplish, minus the need for a private universe.
+# and `--allow-pre` reaches the resolver as the predicate of a `Problem`
+# constraint kind (`prerelease`), forbidding the prereleases of the packages it
+# was not given for -- which is what deleting them used to accomplish, minus the
+# need for a private universe.
 #
 # `allow_pre` is keyed by package uuid with the zero uuid holding the default,
 # exactly as the flag parses it.
 prerelease_exclusion(allow_pre::Dict{UUID,Bool}) =
-    :prerelease => function (uuid::UUID, v::VersionNumber)
+    function (uuid::UUID, v::VersionNumber)
         isempty(v.prerelease) && return false
         return !get(allow_pre, uuid, get(allow_pre, UUID(0), false))
     end
--- a/bin/resolve.jl
+++ b/bin/resolve.jl
@@ -493,9 +493,7 @@ reg = registry_provider(packages;
 # so the struck versions are not in the universe to begin with -- see the provider.)
 const problem = Problem(reqs;
     compat = project_compat,
-    excludes = [
-        prerelease_exclusion(allow_pre),
-    ],
+    prerelease = prerelease_exclusion(allow_pre),
 )
 
 pkg_info = Resolver.pkg_info(reg, problem)
--- a/bin/test/runtests.jl
+++ b/bin/test/runtests.jl
@@ -72,9 +72,8 @@ function make_problem(
 )
     c = Dict{UUID,Any}(compat)
     c[JULIA_UUID] = julia
-    Resolver.Problem(reqs; compat = c, excludes = [
-        prerelease_exclusion(allow_pre),
-    ])
+    Resolver.Problem(reqs;
+        compat = c, prerelease = prerelease_exclusion(allow_pre))
 end
 
 # The old baked path, for the knobs that used to be applied by deleting versions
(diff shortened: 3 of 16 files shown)


StefanKarpinski:
Superseded by #79, which does the same job — one namespace keyed by kind — with a single `Constraint` type instead of three, at +3 lines of `src` against +109, and no measurable performance difference between them.

PR #79, opened 2026-08-11
One map, one constraint type

StefanKarpinski:
An alternative to #78, doing the same job with one type instead of three. Both
are open so they can be compared; close whichever loses.

A `Problem` kept its constraints in three shapes — a compat dict, a pins dict,
and a vector of `kind => predicate` pairs — while every consumer read them as
one namespace keyed by symbol. `exclusion_sources` said so plainly:

```julia
s === nothing || v ∈ s || push!(srcs, :compat)
w === nothing || v == w || push!(srcs, :pin)
for (kind, forbids) in prob.excludes
    forbids(p, v)::Bool && push!(srcs, kind)
end
```

## One type

```julia
struct Constraint{P}
    forbids :: Any                     # (p, v) -> Bool
    named   :: Union{Nothing, Set{P}}  # packages it names, or every package
end
```

What a constraint is *made of* matters only where it is built, so `:compat` and
`:pin` are named in exactly one place in the codebase:

```julia
constraint(::Type{P}, ::Val{:compat}, d) = ...   # a dict → predicate + names
constraint(::Type{P}, ::Val{:pin},    d) = ...
constraint(::Type{P}, ::Val,          f) = Constraint{P}(f, nothing)
```

Every consumer is then a line, and none mentions a kind:

```julia
is_excluded(prob, p, v)     = any(c -> c.forbids(p, v)::Bool, values(prob.constraints))
exclusion_kinds(prob, p, v) = sort!(Symbol[k for (k, c) in prob.constraints if c.forbids(p, v)::Bool])
is_constrained(prob)        = !isempty(prob.constraints)
relaxations(prob)           = Dict(kind => c.named for (kind, c) in prob.constraints)
```

`named` is `nothing` for a constraint that speaks about every package, a
predicate being unable to list them. So lifting a kind entirely is lifting it
for `named(c)`, `relax(c, c.named) === nothing` whatever the shape, and
"lifted entirely" needs no separate spelling. Every kind lifts per package —
including `:compat` and `:pin` — which is what a repair saying "allow
prereleases of X" needs and what the old vocabulary could not express.

## The constructor

```julia
Problem(reqs; compat = Dict(p => spec), pin = Dict(q => v), pre = (p, v) -> ispre(v))
```

Every keyword is a kind and its name is the kind's name; `compat` and `pin`
take a package-keyed dictionary, anything else a predicate, checked once by
`check_constraints`. Two constraints can't share a kind and that needs no check
— a repeated keyword is a syntax error.

`resolve` and `issatisfiable` lose their `compat`/`pins` keywords rather than
slurping: their keyword space already holds `by`, `order` and `diagnose`, and
`order` takes a callable, so a typo like `oder = f` would have become a
predicate kind and quietly constrained the resolve. A constrained resolve takes
a `Problem`.

## Against #78

| | #78 | this |
|---|---|---|
| types | `Compat`, `Pins`, `Predicate` + closed union | `Constraint` |
| methods over them | 9 | 0 (two fields) |
| `src` net | +109 | **+3** |
| total net | +208 | **+99** |

Performance is the same. Measuring `main`, #78 and this branch with one script:
prepare allocations agree to within 0.03% on a constrained query and are
byte-identical on an unconstrained one, and the times sit inside this machine's
noise. `forbids::Any` is a dynamic call where #78 has a union split, and it
does not show up — the closure bodies stay concretely typed because they
capture a concrete `Dict`, and a compat-only query has a small candidate set.

(#78's description previously claimed a 4.5–10% allocation win. That came from a
harness I had not run; it does not reproduce, and #78 has been corrected.)

## What it gives up

A `Problem` no longer remembers *how* a constraint was written — a compat bound
becomes a closure, so the bound cannot be read back out. Nothing in the
codebase needs that: it surfaced only in a test whose reference implementation
reached into `.allowed`, which now recomputes through `is_excluded` on a
single-kind problem instead, and is the better reference for it. But #78 keeps
the capability and this does not, so if a diagnosis ever wants to *show* a
bound rather than name the kind that excluded a version, it would need the
inputs rather than the `Problem`.

### DIFF
--- a/src/Problem.jl
+++ b/src/Problem.jl
@@ -11,27 +11,34 @@ Base.haskey(::EmptyDict, key) = false
 Base.get(::EmptyDict, key, default) = default
 Base.getindex(::EmptyDict, key) = throw(KeyError(key))
 
-# The kinds with no versions to forbid: the shared empty value for a problem's
-# `excludes`, for the same reason `EmptyDict` exists.
-const NoKinds = Pair{Symbol,Any}[]
+# What a constraint is made of matters only where it is built. Everywhere else
+# it answers two questions: does it forbid this version of this package, and
+# which packages does it speak about — `nothing` for one that speaks about every
+# package, a predicate having no way to list them.
+struct Constraint{P}
+    forbids :: Any                     # (p, v) -> Bool
+    named   :: Union{Nothing, Set{P}}  # the packages it names, or every package
+end
 
 """
-    Problem(reqs; compat = ..., pins = ..., excludes = ...)
+    Problem(reqs; compat = ..., pin = ..., \$kind = ...)
 
 A resolution problem: the required packages `reqs`, plus the constraints that say
 which versions are admissible. A `Problem` carries everything that determines
 *satisfiability*; orderings (the `by` priority order, the `order` version rank
 order) are `resolve` parameters instead.
 
-Three kinds of constraint, in two shapes:
+Every keyword is a constraint, and its name is the constraint's *kind*:
 
   * `compat`: per package, the set of allowed versions (queried with `in`).
-  * `pins`: per package, the one version it is held at.
-  * `excludes`: the *admission* knobs — "no prereleases" is the one the resolver's
-    own tooling uses — which say something about versions rather than about
-    particular packages, and so come as `kind => predicate` pairs, where
-    `predicate(p, v)` is true for the versions that source forbids and `kind` is a
-    symbol naming it.
+  * `pin`: per package, the one version it is held at.
+  * anything else: a predicate `(p, v) -> Bool`, true for the versions that kind
+    forbids — "no prereleases" is the one the resolver's own tooling uses. These
+    speak about versions rather than about particular packages, so they reach
+    every package.
+
+Two constraints cannot share a kind, which needs no checking: a repeated
+keyword is a syntax error.
 
 Constraints for packages that don't exist, and constraints that exclude nothing,
 are allowed and have no effect.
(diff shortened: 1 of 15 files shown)


src/Problem.jl
```diff
@@ -11,27 +11,34 @@ Base.haskey(::EmptyDict, key) = false
...
+# which packages does it speak about — `nothing` for one that speaks about every
+# package, a predicate having no way to list them.
+struct Constraint{P}
+    forbids :: Any                     # (p, v) -> Bool
+    named   :: Union{Nothing, Set{P}}  # the packages it names, or every package
+end
 
 """
-    Problem(reqs; compat = ..., pins = ..., excludes = ...)
+    Problem(reqs; compat = ..., pin = ..., kind = ...)
```
StefanKarpinski:
```suggestion
    Problem(reqs; compat = ..., pin = ..., \$kind = ...)
```
The `$` indicates that `kind` is not the literal keyword name.

src/Resolve.jl
```diff
@@ -232,27 +226,11 @@ permutations it induces are computed once however often the result is resolved.
 function relax(
     univ :: Universe{P,V},
     Q    :: Problem{P},
-    drop_reqs    :: SetOrVec{P} = P[],
-    drop_sources :: SetOrVec{Pair{Symbol,Any}} = Pair{Symbol,Any}[];
+    drop_reqs :: SetOrVec{P} = P[],
+    drops     :: AbstractDict{Symbol} = EmptyDict{Symbol,Nothing}();
```
StefanKarpinski:
```suggestion
    drop_constraints     :: AbstractDict{Symbol} = EmptyDict{Symbol,Nothing}();
```
This feels like a more consistent keyword name.

PR #84, opened 2026-08-21
Present minimal fixes as a factored product of independent choices

### LINKED ISSUE #83: Diagnosis: present minimal fixes as a factored product of independent choices
Today an unsatisfiable diagnosis reports "N conflicts" and then **one flat menu of fixes covering all of them at once**. Because that menu enumerates across conflicts it is a *product*: k conflicts of m repairs each give up to mᵏ combinations, truncated at `MAX_FIXES = 32`. The result is long, repetitive, and silently incomplete.

This issue is a handoff: the plan, the measurements behind it, and the approaches that were tried and rejected so they are not re-proposed.

**The branch `diagnose-per-conflict-fixes` (commit `e2f7b87`) implements an earlier version of this plan and is superseded.** It is not wrong so much as built on a decomposition we no longer believe in; see "What changed" at the bottom.

## The plan

Everything below is computed from the MCS family, which the diagnosis already enumerates.

### 1. Enumerate MCSes and take the minimum-cardinality ones

`sat_mcses(sat, [req_lits; pkg_lits])` inside `with_classes_relaxed` + `with_emptied_packages`, exactly as `diagnose` already does. Let `minsize` be the smallest MCS size and `F` the MCSes of that size.

This is cheap: 6–36 MCSes per query in measurement, against 2,000–19,000 solves for even a size-bounded MUS enumeration.

`minsize` is worth reporting on its own — *"the smallest fix changes N things"* — and unlike the current conflict count it is an invariant of the problem rather than an artifact of an algorithm.

### 2. Factor F into a Cartesian product

Two literals belong to the same coordinate iff they **never** co-occur in a member of `F`; literals in different coordinates **always** co-occur. So: build the co-occurrence graph on `∪F`, take connected components of its complement, and verify

- no two literals within a coordinate co-occur,
- every cross-coordinate pair does co-occur,
- `|F| == prod(length, coordinates)`,
- every member of `F` has exactly one literal per coordinate.

If that holds, `F` is exactly the product of the coordinates. Note `#coordinates == minsize` follows: each fix takes one literal per coordinate.

If it does not hold, cover `F` with **rectangles** (a shared core plus a set of alternatives) and take the largest. Greedy over cores of size `minsize-1` was enough in measurement; minimum rectangle cover is hard in general but these families are tiny.

### 3. Give each coordinate a conflict story

A coordinate says *"pick one of these actions"* without saying why. To explain it, take any `M ∈ F`, and for coordinate `i` with chosen element `cᵢ`:

- drop `M \ {cᵢ}` from the assumed literals,
- extract a MUS of what remains.

Minimality of `M` guarantees that MUS contains `cᵢ` and **no other element of `M`** — so each coordinate gets its own conflict, and the conflicts explain the recommended fix rather than being derived independently of it. Verified 2/2 and 2/2 on the two cases where complete MUS enumeration was possible.

That is one MUS extraction per coordinate; coordinates numbered 1–4 in measurement.

### 4. Present it

Each coordinate is a conflict: its story, then its menu of alternatives. A coordinate with a single alternative is a **forced action**, and should read as one rather than as a one-item menu.

**Every combination of choices is a minimum-size MCS by construction**, so every path through the menus is a verified complete fix. No verification pass, no composition check, no "this may reveal further conflicts" warning.

### 5. The qualifier sentences

Two independent conditions:

- **(a)** every MCS is of minimum size — nothing larger exists;
- **(b)** one rectangle covers all of `F` — the product is the whole minimum-size family.

If both hold, the product *is* the complete set of minimal fixes and the report needs **no qualifier**.

- (a) fails, (b) holds → **"Other, larger solutions exist."**
- (b) fails → **"Other solutions exist."**

(When both fail, "Other solutions exist" is the one to print — it is the stronger omission. Flagging that as a reading of the intent rather than something specified.)

Letting the user *explore* the other solutions is out of scope; the sentence is enough for now.

## Measurements

80 failing registry queries, generated by pinning requirements to their newest version and part of their dependency closure to its oldest:

| | |
|---|---|
| (a) every MCS is minimum size | 29/80 (36%) |
| (b) min-size family is one rectangle | 66/80 (83%) |
| both → unqualified report | 23/80 (29%) |
| when (b) fails, largest rectangle covers | median 67% of minimum fixes |
| rows: all MCSes → min-size → coordinates | median 4.5 → 2.0 → 2.0 |

So the qualifier is the *common* case (~70%) and should read as ordinary, not as a warning. Coordinates observed: 1–4, with 1–4 choices each.

Two supporting results:

- Restricting to minimum size **prunes the action space by 46%** (`|UMU|` median 8 → `|U_min|` median 4). Those are constraints that appear in some minimal repair but never in a cheapest one.
- Restricting to minimum size does **not** break the "differs by one action" behaviour — it concentrates it (52% → 67% of same-size pairs). That is expected: the minimum-size MCSes *are* the product of the cheapest choice per conflict. The product is intrinsic; the fix is to present it factored rather than enumerated.

Conflict menus are small independently of all this: mean 2.5 repairs per conflict, and 74% of conflicts offer no choice at all (19% have only "drop a requirement", 55% have exactly one relaxation).

## Rejected — do not re-propose

**The conflict count `k = length(sat_disjoint_muses(...))`.** Heuristic and overstating. It is a greedy loop — extract a MUS, delete *all* its literals, repeat — so disjointness is manufactured rather than discovered, and `sat_mus` is not even stable across differently-warmed solvers (see the determinism note in `UnsatCores.jl`). Three independent measurements say it overstates independence: MUS-graph blocks show a single entangled component where it reports 4–5 conflicts; the MCS family never factors when *all* MCSes are considered (4/45); and `minsize` disagrees with `k` in both directions (7 above, 7 below out of 45). Use `minsize` instead — it is an invariant, and it equals the number of coordinates.

**A free "witnesses agree ⇒ safe to compose" certificate. Measured unsound.** The idea: the diagnosis computes a witness per fix, so if witnesses of different conflicts agree on shared packages their union should solve the combination. It fired on 6 queries; of their 36 combinations only 11 were complete — 25 counterexamples. Agreement on *shared* packages says nothing about a conflict between packages appearing in only one witness each. An earlier smaller run showed 26/26 and looked sound.

**A "union of all repairs" negative test.** Sound by monotonicity but useless: 0 of 37 queries had it come back unsatisfiable.

**Bounded exhaustive combination checking.** Superseded — with the factorization, every combination is a fix by construction, so there is nothing to check.

**Modular decomposition of the conflict graph** (`GraphModularDecomposition.jl`) as preprocessing. Filtering already does the work: DataFrames goes from 321 classes / 97% prime unfiltered to **24 classes and zero conflict edges** filtered. What survives filtering on harder queries is 78–91% prime, and the prime core is flat (Plots: 465 children, 440 of them singletons). Nothing to collapse.

**Block-cut decomposition of the MUS co-occurrence graph.** Real structure — one case had a single compat bound joining 3 blocks — but articulation points are not robust under incomplete MUS enumeration: adding MUSes only merges blocks and destroys cuts, and one case lost its cut vertex when a 15th MUS was found. Any cut-vertex claim needs complete enumeration.

**Note on complexity:** deciding whether a constraint belongs to *some* MUS is Σ₂ᴾ-complete ([Mencía, Kullmann, Ignatiev & Marques-Silva, *On Computing the Union of MUSes*](https://digibuo.uniovi.es/dspace/bitstream/handle/10651/53193/On%20Computing.pdf?sequence=1)), so there is no cheap exact route through MUS structure. Two identities from that literature are worth knowing and were verified here: the **union of all MUSes equals the union of all MCSes** (exact in both cases where complete enumeration was possible), and a constraint is in *every* MUS iff removing it alone makes the problem satisfiable — equivalently, iff it is a singleton MCS.

## Verification

- `test/diagnostics.jl`'s existing oracles still apply: the instance itself for conflicts, and a from-scratch prepare-and-resolve for each fix.
- **Test the factorization by checking every combination resolves** on generated multi-conflict data — that is the property the design rests on, and it is cheap to check exhaustively when coordinates are 1–4.
- Test that the qualifier is **absent** when (a) and (b) both hold, not only that it appears when it should.
- Test a case where (a) fails and (b) holds, and one where (b) fails, and check the two sentences are the right way round.
- **Prove each new test discriminates**: break the thing it tests, confirm failure, restore. In this work an obvious-looking diagnosis test passed with the feature sabotaged out, and only sabotage-testing caught it.

## Traps

- `sat_disjoint_muses` must be computed **before** `with_classes_relaxed` if used at all — inside it the deactivations are lifted and the requirements are satisfiable. (This cost a debugging cycle in a measurement harness.)
- The emptied-package literals from `with_emptied_packages` are **fresh variables outside `VarMap`** — `decode` misattributes them. Map them through the `emptied` dict.
- `diagnose` must leave the instance exactly as the failed query posed it; use the existing `with_*` helpers and nothing else.

## What changed from the previous plan

The earlier plan decomposed by `sat_disjoint_muses`, verified each conflict's repairs against a sub-problem, and warned when composition was not checked. All three fall out: the decomposition comes from the factorization, verification is unnecessary because every combination is a minimum MCS, and the warning is replaced by the two precise sentences above. The `k == 1` special case in that plan was also unsound and is gone.

Caveat on the numbers: 80 synthetic queries against one registry snapshot, from a generator that pins requirements newest and closure packages oldest. The 29%/83%/36% figures are load-bearing for the presentation and would be worth re-measuring on a broader sample.

StefanKarpinski:
Closes #83.

An unsatisfiable diagnosis used to report "N conflicts" and then one flat menu of
fixes covering all of them at once. That menu enumerates *across* conflicts, so
it is a product — k conflicts of m repairs each give up to mᵏ combinations,
truncated at 32. Long, repetitive, and silently incomplete.

This computes the minimal correction sets, keeps the smallest, and **factors
them**. Each coordinate of that product becomes a conflict with its own short
menu, and every path through the menus is a complete fix by construction — so
there is no composition check and no hedged warning about what might still be
broken.

## Before / after

Three requirements, three independent bounds. Before, an 18-item flat product
(4 of 18 shown):

```
Unsatisfiable — 3 conflicts, 18 fixes:

  • no version of PrettyTables is available: 0.1.0–3.4.8 are excluded by your compat
Conflict 2: Plots cannot be satisfied. …
Conflict 3: HTTP cannot be satisfied. …

Verified fixes:
  1. relax your compat on PrettyTables, relax your compat on RecipesBase and
     relax your compat on URIs
     → allows: DataFrames 1.8.2, HTTP 1.11.0, Plots 1.41.6, PrettyTables 3.4.8,
       RecipesBase 1.3.4, URIs 1.7.0
  2. relax your compat on HTTP, relax your compat on PrettyTables and relax
     your compat on RecipesBase
     → allows: DataFrames 1.8.2, HTTP 0.8.19, Plots 1.41.6, …
  … 16 more …
```

After — and with **no qualifier**, because those 18 combinations are exactly the
minimal fixes:

```
Unsatisfiable — 3 conflicts, every one of which has to be fixed:

  • you require DataFrames
  • DataFrames: 0.11.7–0.22.7 are excluded by your compat
  • no version of PrettyTables is available: 0.1.0–3.4.8 are excluded by your compat
  Fix it by any one of:
    1. relax your compat on DataFrames   → allows: DataFrames 0.21.8
    2. relax your compat on PrettyTables → allows: DataFrames 1.8.2, PrettyTables 3.4.8

Conflict 2: Plots cannot be satisfied.
  • no version of RecipesBase is available: 0.4.0–1.3.4 are excluded by your compat
  Fix it by any one of:
    1. relax your compat on RecipesBase  → allows: Plots 1.41.0, RecipesBase 1.3.4
    2. drop requirement Plots

Conflict 3: HTTP cannot be satisfied. …
```

## How it works

1. Enumerate the minimal correction sets, keep the minimum-cardinality ones.
2. Factor that family: two actions share a coordinate iff they never appear in
   the same repair. Verify it really is the product (one action per coordinate,
   counts agree); if not, cover with the largest rectangle.
3. Each coordinate is a conflict. Its story is a minimal explanation taken with
   the other coordinates' choices already settled, so it explains *this* choice.
4. `Diagnosis.fixes` is gone; `Conflict.fixes` is new. `others :: Symbol` says
   what the menus leave out: `:none`, `:larger`, `:some`.

Two qualifier sentences, printed only when earned:

- **"Other, larger solutions exist."** — repairs exist that cost more.
- **"Other solutions exist."** — the menus do not reach every cheapest repair.

When neither applies the product *is* the complete set of minimal fixes, and the
report says nothing extra. Measured on 80 generated registry failures: no
qualifier in 29%, one rectangle covers the cheapest repairs in 83%.

## Why "N conflicts" changed meaning

The old count was `length(sat_disjoint_muses(...))` — a greedy loop that extracts
a conflict, deletes all its literals, and repeats, so disjointness was
manufactured rather than discovered. Three measurements said it overstates
independence; #83 records them. The count is now the number of coordinates,
which equals the minimum number of actions any fix needs.

## Conflicts name every requirement they rescue

Worth calling out because it is subtle. When several requirements fail for the
same reason, any one of them already makes that reason unsatisfiable, so
minimality *forces* the others out of the explanation — the report would name one
arbitrarily and the rest would vanish. A user would fix the bound and never learn
the rest had been broken:

```
Conflict 1: DataFrames and CSV cannot both be satisfied.
  • you require DataFrames
  • you require CSV
  • CSV: 0.3.0–0.9.4 are excluded by your compat
  • DataFrames: 0.11.7–0.22.7 are excluded by your compat
  • no version of Tables is available: 0.1.0–1.13.0 are excluded by your compat
  The only fix: relax your compat on Tables
    → allows: CSV 0.10.16, DataFrames 1.8.2, Tables 1.13.0
```

Every requirement the smallest explanation left out is asked whether it can be
satisfied on its own; the ones that cannot are exactly the ones this fix rescues.

## Testing

The design's core claim — every combination of per-coordinate choices resolves —
is checked directly and from scratch, never by asking the diagnosis: 1550
diagnoses, 2507 combinations, 0 failures, with 232 conflicts naming more than one
requirement. `check_diagnosis` also brute-forces every subset of a chain to
confirm the facts are covered by minimal reasons with no passengers.

Every new test was verified to fail when the thing it tests is broken. Disabling
the rescue pass produces exactly 5 failures; swapping the two qualifier sentences,
forcing `others`, truncating menus, and disabling the forced-action branch each
fail their own assertions.

Full suite green on 1.12.6 and on nightly 1.14.0-DEV; docs build clean.

`sat_disjoint_muses` now has no caller but is left in place — it is a documented,
tested `UnsatCores` API, and #83 rejects it as a decomposition, not as a function.

🤖 Reviewed and pushed with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Diagnostics.jl
+++ b/src/Diagnostics.jl
@@ -208,11 +208,17 @@ end
     Resolver.Diagnostics.Conflict{P,V}(reqs, chain, fixes)
 
 One independent choice the query leaves the user: the requirements `reqs`,
-which cannot hold together, `chain`, a short sequence of
+which cannot all hold, `chain`, a short sequence of
 [`Fact`](@ref Resolver.Diagnostics.Fact)s about the packages that says why, and
 `fixes`, the menu of [`Fix`](@ref Resolver.Diagnostics.Fix)es that settle it.
-Every fact in the chain is load-bearing — take any one of them away and the
-rest of the conflict dissolves.
+
+`reqs` is every requirement this conflict's fixes rescue, which need not be one
+thing that fails: requirements can fail together, or each on its own for a
+reason they have in common, and a conflict gathers both. So the chain is
+unsatisfiable and nothing in it is a passenger — every fact belongs to some
+minimal reason the requirements cannot hold — but it is a *union* of those
+reasons rather than a single one, and taking a fact away need not dissolve all
+of them.
 
 The conflicts of a [`Diagnosis`](@ref) are independent in that each has to be
 fixed and any fix of one goes with any fix of another. Their *stories* may well
@@ -380,8 +386,12 @@ function Base.show(io::IO, ::MIME"text/plain", d::Diagnosis{P,V}) where {P,V}
     for (k, c) in enumerate(d.conflicts)
         println(io)
         subject = list_phrase(String[string(p) for p in c.reqs])
-        println(io, "Conflict $k: $subject cannot be satisfied",
-            length(c.reqs) > 1 ? " together." : ".")
+        # a conflict can gather requirements that fail together and
+        # requirements that each fail on their own, so what is said of them is
+        # what is true of both: the conjunction is what cannot hold
+        println(io, "Conflict $k: $subject cannot ",
+            length(c.reqs) == 1 ? "be satisfied." :
+            length(c.reqs) == 2 ? "both be satisfied." : "all be satisfied.")
         for f in c.chain
             f isa Requirement ?
                 println(io, "  • you require $(f.pkg)") :
@@ -425,6 +435,12 @@ end
 # and what a truncated enumeration costs is not a fix but a claim: the smallest
 # of the repairs found still factor into conflicts and their menus still repair
 # the query, but the report can no longer say that nothing else would.
+#
+# Truncation is deliberately reported as `:some` — the same thing a family that
+# is not a product reports — because the sentence is the same either way: there
+# are repairs the menus do not offer. It says less than it could, in a case that
+# is rare enough (the ceiling is an order of magnitude above what was measured)
+# that a third sentence would be a shape of report nobody has read.
 const MAX_REPAIRS = 256
 
 # The literals a diagnosis assumes, read back as what they say about the
@@ -497,12 +513,28 @@ function with_emptied_packages(body::Function, sat::SAT{P}, vm::VarMap{P}) where
     end
 end
 
-# The story of one conflict: the smallest set of the query's own facts that is
-# still unsatisfiable, drawn from the ones it is asked about. Deletion is
+# The story of one conflict: the query's own facts that this conflict's fix is
+# what rescues, and why each of them needs rescuing.
+#
+# The smallest unsatisfiable set of them is where it starts. Deletion is
 # attempted in the order given, so putting the requirements first drops the ones
-# the story does not need and keeps the facts the user can act on. The answer is
-# a subsequence of what was asked, so the requirements come back in the
-# problem's requirement order and the emptied packages in package order.
+# that set does not need and keeps the facts the user can act on.
+#
+# But "does not need" is the trap. Several requirements can fail for one and the
+# same reason, and one of them is enough to make that reason a conflict — so the
+# smallest set names one of them, arbitrarily, and the others go unmentioned
+# though this conflict's fix is what rescues them too. A user would fix the
+# bound and never learn the rest had been broken. So every requirement the set
+# left out is asked whether it can be satisfied on its own against the packages
+# as this conflict finds them; the ones that cannot are exactly the ones this
+# fix rescues, since the fixes of the other conflicts are already in force here
+# and carrying this one out satisfies the query. Each of those brings its own
+# smallest explanation, and the chain is all of them at once.
+#
+# One `sat_mus` call per requirement, which answers both questions: empty when
+# the requirement is satisfiable, and its explanation when it is not. What comes
+# back is a subsequence of what was asked, so the requirements are in the
+# problem's requirement order and the packages in package order.
 function conflict_story(
     sat     :: SAT{P,V},
     prob    :: Problem{P},
@@ -510,11 +542,17 @@ function conflict_story(
     emptied :: Dict{Int,P}, # per package literal, the package
     cands   :: Vector{Int},
 ) where {P,V}
-    mus = sat_mus(sat, cands)
+    named = Set{Int}(sat_mus(sat, cands))
+    pkgs = Int[l for l in cands if haskey(emptied, l)]
+    for l in cands
+        (haskey(emptied, l) || l in named) && continue
+        union!(named, sat_mus(sat, Int[l; pkgs]))
+    end
     reqs = P[]
     chain = Fact[]
     avails = Fact[]
-    for l in mus
+    for l in cands
+        l in named || continue
         p = get(emptied, l, nothing)
         if p === nothing
             q, _ = decode(vm, l)
--- a/src/Diagnostics.jl
+++ b/src/Diagnostics.jl
@@ -304,7 +304,7 @@ end
 # An availability fact as one line. The fact runs over the package's whole
 # version list, so every maximal run of versions excluded by the same kinds is a
 # range and gets one clause, and the runs it admits are simply not mentioned;
-# the clauses after the first drop the verb, as an English list of them would.
+# the clauses after the first drop "excluded", as an English list of them would.
 function availability_phrase(f::Availability)
     members, excluded = f.members, f.excluded
     clauses = String[]
@@ -315,10 +315,9 @@ function availability_phrase(f::Availability)
             j += 1
         end
         if !isempty(excluded[i])
-            verb = i == j ? "is" : "are"
             by = kinds_phrase(excluded[i])
             push!(clauses, isempty(clauses) ?
-                "$(range_phrase(members, i, j)) $verb excluded by $by" :
+                "$(range_phrase(members, i, j)) excluded by $by" :
                 "$(range_phrase(members, i, j)) by $by")
         end
         i = j + 1
--- a/src/Diagnostics.jl
+++ b/src/Diagnostics.jl
@@ -28,7 +28,7 @@ export Diagnosis, Conflict, Fix, Action, Fact, Requirement, Availability,
 
 using ..Resolver: SAT, Problem, Universe, relax, resolve, installed_lit,
     with_temp_clauses, with_classes_relaxed, sat_new_variable, sat_add_var,
-    sat_add, exclusion_kinds, class_exclusions
+    sat_add, sat_solve, exclusion_kinds, class_exclusions
 using ..UnsatCores: sat_mus, sat_mcses
 
 # What an unsatisfiable resolve says, and how it is found.
@@ -426,14 +426,17 @@ end
 
 ## the diagnosis
 
-# How many repairs to enumerate before settling for the ones already found.
-# What is enumerated over is the requirements and the packages this query left
-# short, of which there are few — 6 to 36 repairs on the registry queries this
-# was measured on — so it is a ceiling rather than the usual stopping point.
+# How many of the smallest repairs to enumerate before settling for the ones
+# already found. What is enumerated over is the requirements and the packages
+# this query left short, of which there are few, and only the smallest repairs
+# among them — a median of 2 of them on the registry queries this was measured
+# on — so it is a ceiling rather than the usual stopping point.
 # Independent conflicts multiply, though, so there can be exponentially many,
-# and what a truncated enumeration costs is not a fix but a claim: the smallest
-# of the repairs found still factor into conflicts and their menus still repair
-# the query, but the report can no longer say that nothing else would.
+# and what a truncated enumeration costs is not a fix but a claim: the repairs
+# found still factor into conflicts and their menus still repair the query, but
+# the report can no longer say that nothing else would, nor ask whether
+# anything larger exists — that question is put by ruling every one of the
+# smallest repairs out, so it needs all of them.
 #
 # Truncation is deliberately reported as `:some` — the same thing a family that
 # is not a product reports — because the sentence is the same either way: there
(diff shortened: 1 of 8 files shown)


StefanKarpinski:
Added `bd7a663`, which makes the `others` field exact instead of inferred.

**The problem it fixes.** The repair enumeration was capped, and on truncation `others` fell back to `:some` — "Other solutions exist" — because it could no longer tell whether the repairs it had found were all of them. That is honest but vague, and it degrades exactly where the report is most complicated.

**What it does.** Two changes, both inside `with_temp_clauses`; `UnsatCores.jl` is untouched, since it is deliberately assumption-only.

- Enumerate with a **cardinality bound**. A repair is the set of candidate literals a model leaves unsatisfied, so "at most k" is an at-most-k constraint over their negations (Sinz sequential counter). The bound rises from 1 until the instance gives way, and the enumeration then runs at that bound — so the search never pays for repairs larger than the ones it will present.
- Decide "does anything larger exist" in **one solve**. Given every minimum repair, add one clause per repair — the plain disjunction of its literals — and solve. SAT means some model avoids every minimum repair, so a larger one exists. UNSAT means every correction set contains a minimum repair, and minimality forces every repair to be one.

**What it buys**, on a shape of *g* independent conflicts each fixable by one relaxation:

| | g=2 | g=4 | g=6 | g=8 | g=10 |
|---|---|---|---|---|---|
| before | `:larger` | `:some` | `:some` | `:some` | `:some` |
| after | `:larger` | `:larger` | `:larger` | `:larger` | `:larger` |

From g=4 the old enumeration hits its ceiling and gives the vaguer sentence; the new one answers correctly at every size.

**Not a speed change.** On registry queries every repair is already minimum size, so the bound has nothing to skip and timings are unchanged. The enumeration step alone is 2–4× faster on synthetic shapes, but `diagnose` is dominated by the per-fix resolves and per-conflict MUS extraction, so it does not show up end to end. This is worth having for exactness, not performance.

One implementation note worth recording: `sat_new_variable` is permanent — picosat's max var never shrinks — so building a fresh counter per bound leaves dead variables that every later solve, including the per-fix `resolve` calls, pays to decide. That alone doubled `diagnose`'s time until the counting variables were pooled across the search, which is why `add_at_most!` takes a `counts` store rather than allocating its own.

Verified: inverting the blocking clause's polarity fails the suite (3 assertions), as does making the bound a no-op. Full suite green on 1.12.6 and nightly 1.14.0-DEV; docs build clean.

src/Diagnostics.jl
```diff
@@ -322,59 +352,208 @@ end
...
     print(io, "Diagnosis: ",
         count_phrase(length(d.conflicts), "conflict", "conflicts"), ", ",
-        count_phrase(length(d.fixes), "fix", "fixes"))
+        count_phrase(nfixes(d), "fix", "fixes"))
+end
+
+# What the menus leave out, once the reader has made a choice in each of them.
+# Nothing to say when they leave out nothing.
+others_phrase(others::Symbol) =
+    others === :larger ? "Other, larger solutions exist." :
```
StefanKarpinski:
```suggestion
    others === :larger ? "Larger solutions also exist." :
```

src/Diagnostics.jl
```diff
@@ -322,59 +352,208 @@ end
...
+                         "Other solutions exist."
```
StefanKarpinski:
```suggestion
                         "Other solutions also exist."
```

src/Diagnostics.jl
```diff
@@ -322,59 +352,208 @@ end
...
+                    if haskey(fix.solution, p)]
+    isempty(allows) || print_wrapped(io, join(allows, ", "), lead, rest)
 end
 
 function Base.show(io::IO, ::MIME"text/plain", d::Diagnosis{P,V}) where {P,V}
     println(io, "Unsatisfiable — ",
-        count_phrase(length(d.conflicts), "conflict", "conflicts"), ", ",
-        count_phrase(length(d.fixes), "fix", "fixes"), ":")
+        count_phrase(length(d.conflicts), "conflict", "conflicts"),
+        length(d.conflicts) > 1 ? ", every one of which has to be fixed:" : ":")
```
StefanKarpinski:
```suggestion
        length(d.conflicts) > 1 ? ", each of which must be fixed:" : ":")
```

src/Diagnostics.jl
```diff
@@ -446,25 +625,46 @@ function with_emptied_packages(body::Function, sat::SAT{P}, vm::VarMap{P}) where
...
-    prob     :: Problem{P},
-    vm       :: VarMap{P},
-    cluster  :: Vector{Int},
-    emptied  :: Dict{Int,P}, # per package literal, the package
-    pkg_lits :: Vector{Int},
+    sat     :: SAT{P,V},
+    prob    :: Problem{P},
+    vm      :: VarMap{P},
+    emptied :: Dict{Int,P}, # per package literal, the package
+    cands   :: Vector{Int},
```
StefanKarpinski: I assume `cands` is "candidates"? Should either be commented or spelled out

src/Diagnostics.jl
```diff
@@ -489,34 +693,109 @@ function correction_actions(
...
+    choices = push!(Vector{Int}[Int[k] for k in core], alts)
+    return choices, length(alts) == length(repairs)
+end
+
+# Presentation order for the choices and the alternatives within them, given the
+# order the repairs were asked for in. Within a choice, relaxing a constraint
+# comes before dropping a requirement, which is the same preference the
+# candidate order states by putting the requirements first; the choices
+# themselves go in the order their earliest alternative was offered in.
+function order_choices!(choices::Vector{Vector{Int}}, cands::Vector{Int},
```
StefanKarpinski: Weird formatting that doesn't match the rest of the code. If all args fit on one line, do that, otherwise one arg per line on its own.

StefanKarpinski (commented):
Seems like there should be some diagnostics examples in the docs too.

PR #110, opened 2026-09-09
Fix stale stdlib snapshot, and notice the next one (#103)

### LINKED ISSUE #103: HistoricalStdlibVersions v2.0.2
I am facing an issue when using julia-downgrade-compat and I think it originates from Resolver.jl, see https://github.com/hpsc-lab/ObliviousOffload.jl/issues/45.

HistoricalStdlibVersions is pinned to v2.0.2: https://github.com/StefanKarpinski/Resolver.jl/blob/ea359e31e96a300eb6400d68249aec856c21d7cc/bin/Manifest-v1.9.toml#L31-L33

When running Resolver.jl for julia v1.12.0 and OpenSSL_CLI_jll v3.5.1, this leads to the following conflict:

```
Conflict 1: OpenSSL_CLI_jll cannot be satisfied.
    • you require OpenSSL_CLI_jll
    • your compat and your prerelease setting leaves julia at 1.12.0–1.12.7
    • OpenSSL_CLI_jll 3.5.1+0 requires OpenSSL_jll at ≥ 3.5.7+0
    • OpenSSL_CLI_jll 3.5.1+0 requires julia at ≥ 1.6.0
    • julia 1.12.0–1.12.7 works with OpenSSL_jll only at 3.0.15+0–3.0.15+3
    Fix it by any one of:
      1. relax your compat on julia
         → allows: OpenSSL_CLI_jll 3.5.1+0, OpenSSL_jll 3.5.8+0, julia 1.10.0
      2. drop requirement OpenSSL_CLI_jll
         → allows: OpenSSL_jll 3.0.15+3, julia 1.12.0
```

More recent versions of HistoricalStdlibVersions have OpenSSL_jll 3.5.1 for Julia v1.12.0.

StefanKarpinski:
Fixes #103.

## The bug

`bin/`'s manifests pinned `HistoricalStdlibVersions` at 2.0.2, whose data predates Julia 1.12. The resolver therefore believed 1.12.x bundles `OpenSSL_jll` 3.0.15, when 1.12.0 ships 3.5.1 and 1.12.7 ships 3.5.6 — so `OpenSSL_CLI_jll` would not resolve on 1.12 at all. This reached users through [`julia-downgrade-compat`](https://github.com/julia-actions/julia-downgrade-compat), which uses `bin/resolve.jl`.

Both halves of the reported error trace to that one pin, including the half that looked like a separate bug:

> `OpenSSL_CLI_jll 3.5.1+0 requires OpenSSL_jll at ≥ 3.5.7+0`

@Tom-Finke was right that the registry compat is really `~3.5.1`. The chain renders the requirement *after* filtering, and the bogus 3.0.15 pin had removed `OpenSSL_jll` 3.5.1+0 through 3.5.6+0 from the universe, leaving only ≥ 3.5.7+0 standing. Correcting the pin retires both complaints. 2.0.2 was five releases behind, and the same drift misdated eleven other stdlibs on 1.12.

The reported case now resolves as expected — `OpenSSL_CLI_jll` 3.5.1+0 with `OpenSSL_jll` 3.5.6+0 on julia 1.12.7 — identically under 1.9, 1.10, 1.11 and 1.12.

## Noticing it next time

A stale pin does not fail, it answers wrongly: a Julia newer than the pin is not *missing* from the snapshot, it is described by an older stanza. The resolver keeps answering, about a Julia that never existed, and the wrong answer is shaped exactly like a real conflict. Nothing in the suite could tell the difference, because nothing knew the right answer.

The running Julia does — every stdlib ships the `Project.toml` it was built from. Checking the snapshot against `Sys.STDLIB` costs a directory listing and no network. On the pin as it stood it names twelve drifted stdlibs on 1.12, `OpenSSL_jll` among them.

What it deliberately does **not** do is ask whether a newer `HistoricalStdlibVersions` exists. That would be a moving target — red on the package's release schedule rather than on anything in the tree — and a check that fails for reasons the tree cannot fix gets disabled. This one fails only when the pin actually misdescribes a Julia that is present.

Two findings shaped the design:

- **Nightly can never pass it.** The snapshot's `1.14.0` stanza and a nightly build are two points on a moving branch; they disagree about eighteen stdlibs and neither is wrong. CI runs the suite on nightly, so an unguarded testset would have been permanently red. Prereleases are skipped. (1.13.0-rc4 happens to agree with its stanza, but an rc can drift from one through nobody's fault, so the line is drawn at released versions.)
- **A missing manifest doesn't fail, it falls back.** A Julia with no `Manifest-vX.Y.toml` reads `Manifest.toml` — the 1.9 symlink — and would overwrite 1.9's pins with its own. So `--check` reports a released Julia that has no manifest, and `update_manifests.jl` creates one (removing the placeholder again if that Julia fails, since an empty manifest would be committed and then silently used as one).

`.github/workflows/stdlibs.yml` sweeps 1.9–1.12 with the standalone check, instantiating each manifest first, so a manifest that stops resolving on its own Julia fails too. Weekly cron, since both failure modes arrive with time rather than with a commit.

## A pre-existing bug this turned up

The `bin/` suite had never been run on 1.9 or 1.10 — it isn't part of CI, which tests the package suite on release and nightly. It errored 7 times on both. Through Julia 1.10 an `IOBuffer` that has served as a subprocess's stderr comes back non-writeable, so `resolve_versions`'s take-and-put-back threw on exactly the path the unsatisfiable-project tests take. Unrelated to this fix, so it's a separate commit. With it, the suite is 178/178 on all four.

## Commits

| | |
|---|---|
| `ee7f08f` | the fix — `HistoricalStdlibVersions` 2.0.2 → 2.0.7 |
| `f716991` | the pre-existing 1.9/1.10 test-helper bug |
| `b9c769a` | the checks, the refresh script and the workflow |
| `7186372` | the manifest refresh |

**The last commit is separable and I'd suggest reviewing it on its own terms.** Running `update_manifests.jl` produces more than #103 needs: JSON 0.21.4 → 1.8.0 (a rewrite, bringing Parsers 3.0 and StructUtils), plus `libpicosat_jll` 965.0.0+0 → 965.0.2+0 and each host Julia's own stdlib versions. All four Julias pass with it, but a JSON major version and a new SAT binary are worth taking deliberately rather than as freight on a bug fix. Dropping it leaves #103 fixed and the checks working.

## Testing

- `bin/` suite: 178/178 on each of 1.9, 1.10, 1.11, 1.12
- package suite: passes
- both checks have negative controls — the stale 2.0.2 pin produces 12 named drifts on 1.12, and a removed `Manifest-v1.11.toml` produces `no manifest for: 1.11`

One thing left for maintenance: the `snapshot` matrix is spelled out rather than derived, so a new Julia minor needs adding there too. The `--check` failure message says so explicitly, since otherwise the matrix would keep passing on the versions it *does* list — the same silent-failure shape this change exists to avoid.

---

### DIFF
--- a/.github/workflows/stdlibs.yml
+++ b/.github/workflows/stdlibs.yml
@@ -1,0 +1,75 @@
+# Check bin/'s pinned stdlib snapshot against every Julia it claims to support.
+#
+# bin/ pins HistoricalStdlibVersions, whose data is a snapshot of what each
+# Julia release bundles. A pin left behind by new Julia releases keeps
+# resolving -- it just resolves against a Julia that never existed, and the
+# wrong answer is shaped exactly like a real conflict (issue #103). The suite in
+# ci.yml runs this same check, but only under the Julia it runs on, so the older
+# manifests would go unchecked. This job covers the rest of the range.
+#
+# It also instantiates each manifest before checking it, so a manifest that no
+# longer resolves on its own Julia fails here too.
+#
+# Only released Julias: bin/check_stdlibs.jl skips prereleases, since a
+# development build and the snapshot's stanza for it are two different points on
+# a moving branch and disagree for no reason anyone could fix.
+name: stdlibs
+on:
+  pull_request:
+    paths:
+      - 'bin/**'
+      - '.github/workflows/stdlibs.yml'
+  push:
+    branches:
+      - main
+    paths:
+      - 'bin/**'
+      - '.github/workflows/stdlibs.yml'
+  # the pin can go stale with nothing here changing, so also check on a timer
+  schedule:
+    - cron: '0 6 * * 1'
+  workflow_dispatch:
+jobs:
+  snapshot:
+    name: Julia ${{ matrix.version }} stdlib snapshot
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # Every minor bin/Project.toml's `julia` compat admits. Spelled out
+        # rather than derived: it changes about once a year, and the
+        # `manifests` job below fails when a released Julia is missing from
+        # bin/, which is the same moment this list needs a new entry.
+        version: ['1.9', '1.10', '1.11', '1.12']
+    steps:
+      - uses: actions/checkout@v7
+      - uses: julia-actions/setup-julia@v3
+        with:
+          version: ${{ matrix.version }}
+      - name: Instantiate bin/
+        run: julia --project=bin -e 'using Pkg; Pkg.instantiate()'
+      - name: Check the stdlib snapshot
+        run: julia --project=bin bin/check_stdlibs.jl
+
+  # A Julia minor with no manifest of its own does not fail -- it falls back to
+  # another Julia's pins -- so nothing above would notice a new release going
+  # unsupported. This is the only check here that depends on what Julia has
+  # shipped rather than on what is installed, so it runs once, on latest.
+  manifests:
+    name: A manifest per released Julia
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - uses: julia-actions/setup-julia@v3
+        with:
+          version: '1'
+      - name: Instantiate bin/
+        run: julia --project=bin -e 'using Pkg; Pkg.instantiate()'
+      # bin/julia_versions.json is a committed cache refreshed when it is over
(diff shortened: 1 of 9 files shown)


Tom-Finke (approved):
@StefanKarpinski thank you for addressing this so quickly. 
This fixes the problem for us, see https://github.com/Tom-Finke/ObliviousOffload.jl/pull/3.

PR #111, opened 2026-09-09
Make @alldeps mean both halves (#102)

### LINKED ISSUE #102: `--min=@alldeps` does not include weakdeps despite the help text
The help text for `bin/resolve.jl` says:

```
  * @alldeps for packages in both the [deps] and [weakdeps] sections
```

but `--min=@alldeps` only minimises the `[deps]` packages. In `parse_packages` (`bin/resolve.jl`, at `ea359e3`):

```julia
function parse_packages(str::AbstractString)
    pkgs = UUID[]
    for item in split(str, ',')
        if item in ("@deps", "@alldeps")
            union!(pkgs, project_deps)
            continue
        end
        if item in ("@weakdeps", "@alldeps")
            union!(pkgs, project_weakdeps)
            continue
        end
```

`@alldeps` matches the first branch and `continue`s, so the `@weakdeps` branch is never reached for it. This has been the case since the selectors were introduced (`a377b93`).

Observed on a project with `[weakdeps]` (DyadModelOptimizer, weakdeps Makie / Distributions / EnzymeCore):

| command | Makie | Distributions | EnzymeCore |
|---|---|---|---|
| `--min=@alldeps --extra-deps=@weakdeps` | 0.24.13 (max) | 0.25.131 (max) | 0.8.21 (max) |
| `--min=@alldeps --min=@weakdeps --extra-deps=@weakdeps` | 0.22.0 | 0.25.123 | 0.8.8 |

Before proposing a fix I'd like to check the intended behaviour:

1. Should `@alldeps` expand to `[deps] ∪ [weakdeps]` as the help text says (i.e. drop the `continue` in the first branch, or fold both into one `if`)? Or is the current behaviour intended and the help text is what should change?
2. Related: is `--min=@weakdeps` meant to have any effect without `--extra-deps=@weakdeps`? Weakdeps are not required by default, so minimising them only matters once they are in the graph; if that is intended, it might be worth a note in the help text.

Happy to send a PR either way.

StefanKarpinski:
Fixes #102.

@Tom-Finke's diagnosis was exactly right: `parse_packages` matches `@alldeps` in the `@deps` branch and `continue`s, so the `@weakdeps` branch it also matches is never reached.

## It is wider than the ordering options

`parse_packages` backs *every* option that takes `<pkgs>`, so `@alldeps` was a plain synonym for `@deps` throughout — not just in `--min`. The case that bites hardest isn't an ordering option at all:

| | before | after |
|---|---|---|
| `--extra-deps=@alldeps` | weak deps not pulled in **at all** | pulled in |
| `--min=@alldeps` | weak deps left at their default maximum | minimized |

For `--extra-deps` the difference is *whether* a package is resolved rather than which version of it — a silently smaller dependency graph.

## Answering the two questions

**1. Should `@alldeps` expand to `[deps] ∪ [weakdeps]`?** Yes — the help text has said so since the selectors were introduced, so the text is right and the code was wrong. Both branches now run for `@alldeps`, with one `continue` covering all three selectors.

**2. Is `--min=@weakdeps` meaningful without `--extra-deps=@weakdeps`?** It isn't a special case. Ordering options are preferences over whatever gets resolved, so naming an unresolved package is inert — the same for a weak dependency as for any unrelated package. It's not dead by rule either: if some other package in the graph depends on it, it gets resolved and the preference applies. But weak dependencies are the one place where "named but not required" is the *default* state, which is where people will trip, so the help text now says it:

> Weak dependencies are not required by default, so a selector naming one only bites on a resolve that pulls it in — through `--extra-deps`, or through some other package that depends on it.

## Testing

The regression test states the fix as an identity — naming the union does what naming both halves does — rather than pinning version numbers the registry will move under. On its own that would prove nothing, since *agreeing* is precisely what the bug did, so it is paired with an inequality showing the weak dependency actually moved.

Negative control, with only the control flow reverted and the test left in place: 3 of the 6 assertions fail, `Compat` stuck at 3.47.0 where the fix gives 1.0.0. Suite is 180/180.

## One drive-by, easy to drop

The parser reported bad input as `Invalid value for --prioritize` whatever option it came from — which is every option taking `<pkgs>`. It now says `Invalid package selector`. Separable from the fix if you'd rather not carry it.

---

### DIFF
--- a/bin/resolve.jl
+++ b/bin/resolve.jl
@@ -47,6 +47,10 @@ Wherever <pkgs> appears you can specify a comma-separated list of:
   * @deps for packages in the [deps] section of Project.toml
   * @weakdeps for packages in the [weakdeps] section of Project.toml
   * @alldeps for packages in both the [deps] and [weakdeps] sections
+
+Weak dependencies are not required by default, so a selector naming one
+only triggers on a resolve that pulls it in through --extra-deps, or
+through some other package that depends on it.
 """
 
 parse_opts!(ARGS, split("""
@@ -287,12 +291,10 @@ project_compat[JULIA_UUID] = julia_versions
 function parse_packages(str::AbstractString)
     pkgs = UUID[]
     for item in split(str, ',')
-        if item in ("@deps", "@alldeps")
-            union!(pkgs, project_deps)
-            continue
-        end
-        if item in ("@weakdeps", "@alldeps")
-            union!(pkgs, project_weakdeps)
+        # @alldeps is both halves, so neither branch may return early on it
+        if item in ("@deps", "@weakdeps", "@alldeps")
+            item in ("@deps", "@alldeps") && union!(pkgs, project_deps)
+            item in ("@weakdeps", "@alldeps") && union!(pkgs, project_weakdeps)
             continue
         end
         uuid = tryparse(UUID, item)
@@ -305,7 +307,8 @@ function parse_packages(str::AbstractString)
                 uuid = only(uuids)
             end
         end
-        uuid isa UUID || usage("Invalid value for --prioritize: $item")
+        # any option taking <pkgs> lands here, so do not name one of them
+        uuid isa UUID || usage("Invalid package selector: $item")
         uuid ∉ pkgs && push!(pkgs, uuid)
     end
     return pkgs
--- a/bin/test/runtests.jl
+++ b/bin/test/runtests.jl
@@ -118,16 +118,23 @@ const RESOLVE_JL = normpath(joinpath(@__DIR__, "..", "resolve.jl"))
 const BIN_PROJECT = normpath(joinpath(@__DIR__, ".."))
 
 function resolve_versions(
-    compat :: AbstractString,
-    flags  :: Vector{String} = String[];
-    deps   :: Vector{Pair{String,UUID}} = ["JSON" => JSON],
-    err    :: IOBuffer = IOBuffer(), # the script's stderr, for tests that read it
+    compat   :: AbstractString,
+    flags    :: Vector{String} = String[];
+    deps     :: Vector{Pair{String,UUID}} = ["JSON" => JSON],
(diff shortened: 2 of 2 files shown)


bin/resolve.jl
```diff
@@ -47,6 +47,10 @@ Wherever <pkgs> appears you can specify a comma-separated list of:
   * @deps for packages in the [deps] section of Project.toml
   * @weakdeps for packages in the [weakdeps] section of Project.toml
   * @alldeps for packages in both the [deps] and [weakdeps] sections
+
+Weak dependencies are not required by default, so a selector naming one
+only bites on a resolve that pulls it in -- through --extra-deps, or
```
StefanKarpinski:
```suggestion
only triggers on a resolve that pulls it in through --extra-deps, or
```
Bites is a weird AI-ism. So is the unnecessary em dash.

bin/test/runtests.jl
```diff
@@ -316,6 +323,36 @@ end
               Resolver.resolve(info, prob)[JULIA_UUID]
     end
 
+    # `@alldeps` is documented as "packages in both the [deps] and [weakdeps]
+    # sections", but it matched the `@deps` branch of the selector parser and
+    # stopped there -- so it was a plain synonym for `@deps` in every option
+    # that takes a package selector, not just the ordering ones (#102).
```
StefanKarpinski: Don't give a history of the code in the comments. Explain only what the code does now and why and only if necessary; if it's clear what it does and why, don't comment.

bin/resolve.jl
StefanKarpinski: Taken as suggested.

bin/test/runtests.jl
StefanKarpinski: Rewritten to say what the test checks and why it is shaped that way, with the bug narrative dropped. Also applied it to the parser comment in bin/resolve.jl, which carried the same history — it is now just the one line saying why neither branch returns early.

PR #123, opened 2026-09-17
Print the report in color where the stream takes it

KristofferC:
A report names a handful of packages over and over, and a reader following a chain is tracking which line speaks of which. When `show` is given a stream that takes color (`get(io, :color, false)`), each package now prints in a color of its own and a version in the color of the package it belongs to, the way Pkg's maxsum resolve log does. Without color the page is byte-for-byte what it was.

- Color is added to the finished page rather than while it is composed: the page is wrapped and checked as plain text, and a colored page differs from it by escapes alone. Nothing in the phrase code changes.
- The palette and the choice of color are Pkg's own (`CONFLICT_COLORS = [1:6; 10:14]` picked by `hash(name)`), so a package reads in the same color in both resolvers' output.
- A version takes the color of the package named before it in its clause, which is where every sentence on the page puts one ("DataFrames 1.8.2", "restricts P to p1"). After the semicolon of an upstream sentence no package precedes the version, and it is left plain rather than guessed at.
- Names and versions are matched whole: `JSON` is not colored inside `JSON3`, nor `1.2` inside `1.2.3`.

## Before / after

Two conflicts against the General registry (`DataFrames = "1"` with `PrettyTables = "0.9"`, `Plots = "1.40"` with `RecipesPipeline = "0.5"`):

| Before | After |
| --- | --- |
| plain report, two conflicts | colored report, two conflicts |

One conflict:

| Before | After |
| --- | --- |
| plain report, one conflict | colored report, one conflict |

Pkg builds the report with `sprint` and so still gets the plain page; passing `context = stderr` there, as its maxsum `logstr` does, would turn color on. That is a one-line follow-up on the Pkg side.


StefanKarpinski:
I see that this is about color coding the same name consistently, but it took me quite a while to understand that from just looking at the output. The use of red and yellow especially suggests that the colors have something to do with being problematic, whereas the colors don't actually mean that.

KristofferC:
It's basically the same color coding strategy as what is in the current resolver, I find it somewhat useful but agree that it is easy to assign some semantic meaning to them.

StefanKarpinski:
How about we hold off on this and see if it's missed significantly?

